PR #2, opened 2023-05-10
Publish workflow

phackstock:
This PR adds an automated workflow for pushing to (test)-pypi.
There are two triggers for the publish workflow:

1. Publish to pypi when a release is created on GitHub
2. Publish to test-pypi when a tag with `v*` is pushed

I've set configured it so that the two events are **mutually** exclusive. To avoid double running the workflow since the release on GitHub goes along with pushing a tag.

I have also included the poetry dynamic versioning plugin (equivalent to setuptools_scm).

Any input for changes to the README or DEVELOPING are welcome.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,14 +12,14 @@ jobs:
       # publish to pypi
       - if: github.event_name == 'release'
         name: Build and publish to pypi
-        uses: JRubics/poetry-publish@v1.16
+        uses: JRubics/poetry-publish@v1
         with:
           pypi_token: ${{ secrets.PYPI_TOKEN }}
           plugins: "poetry-dynamic-versioning-plugin"
       # publish to testpypi
       - if: github.event_name != 'release'
         name: Build and publish to testpypi
-        uses: JRubics/poetry-publish@v1.16
+        uses: JRubics/poetry-publish@v1
         with:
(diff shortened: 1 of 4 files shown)


glatterf42:
Agreed, a comment in the docs about the version number and its automaticity would be useful. I'd approve it afterwards, but @meksor's opinion might be more important here.

meksor:
Hey sorry for leaving this untouched for so long. In anticipation of the 0.2.0 release: Is there any potential way to sync this version number with the one in ixmp4/__init__.py?

https://github.com/iiasa/ixmp4/blob/main/ixmp4/__init__.py

phackstock:
Plugging shamelessly from stackoverflow (https://stackoverflow.com/questions/67085041/how-to-specify-version-in-only-one-place-when-using-pyproject-toml), this should do it:

```python
import importlib.metadata

__version__ = importlib.metadata.version("mypackage")
```

the advantage of this is that it does not need to be manually udpated.

meksor:
oh, perfect! Thanks!

meksor (approved):

meksor:
So your suggestion is in #4. I removed the version test as that will sometimes fail between releases (i think).
Merge when you are ready and i will integrate the changes into #4.

phackstock:
@meksor just merged, so you can integrate the changes into #4 now.

PR #6, opened 2023-08-10
Fix publish workflow

phackstock:
I've been playing around with this for a little while now. 
1. Issue was related to JRubics/poetry-publish (details: https://github.com/JRubics/poetry-publish/issues/39) solved by downgrading from v1.17 -> v1.16
2. Issue was that the `pg_config` command was not found which is apparently required for building `psycopg2`. 
I tried the following fixes:
  * Bumping the version of psycopg2 (didn't work)
  * Installing `libpq-dev` (didn't work and was already installed anyway) 
  * Switch `psycopg2` to `psycopg2-binary` which is the pre-built version (apparently not recommended in production...)

The third option worked though (https://github.com/iiasa/ixmp4/actions/runs/5821805161, failed because of the version number as it was not triggered from a tag) so that sure what to do...


phackstock:
Seems like my update to poety.lock broke some of the linting and formatting checks, reverting.

danielhuppmann:
I remember that psycopg2 gives me a headache every time I install ixmp4 in a new environment...

phackstock:
Played around a bit more and it seems that using `psycopg2-binary` is the only way to run the publish workflow...

phackstock:
The strange thing is that it seems all the other workflows are able to install the package just fine ... 
I'll try that.

phackstock:
Looking good now https://test.pypi.org/project/ixmp4/0.2.0rc5/

danielhuppmann:
🎉

phackstock:
I think I figured out what was happening yesterday and why I couldn't get the workflow to work with JRubics/poetry-publish. 
The strange part was that I got an error that `pg_config` was missing. When I than ran `sudo apt-get install libpq-dev` I got a message saying that it was already installed and that didn't fix the error. 
Now I've realized that JRubics/poetry-publish runs in it's own docker container inside the GitHub action. `pg_config` was just installed on the host system, the `ubuntu-latest` runner from GitHub but not the container of the action.
That's why when I changed the publish workflow to not use JRubics/poetry-publish, everything is not working fine.

phackstock:
Just tested the install from testpypi, looks all good. The only reason that the publish workflow is failing here is because I pushed the `v0.2.0rc5` tag in parallel and the workflow triggered by the tag ran faster (https://github.com/iiasa/ixmp4/actions/runs/5823260950).
All good to merge from my side and release ixmp4 0.2.0.

danielhuppmann:
> I pushed the v0.2.0rc5 tag in parallel and the workflow triggered by the tag ran faster

Bit why would the push of a commit trigger an upload to test-pypi at all? Only the tags should have any impact on pypi and test-pypi.

phackstock:
@danielhuppmann good question, just took a look and I cannot tell you, from looking at https://github.com/iiasa/ixmp4/actions/workflows/publish.yaml it looks like both of the runs for `v0.2.0rc5` were triggered by the exact same event....
No idea why that ran twice.

phackstock:
Regardless of what caused the double run though, the package did end up being uploaded correctly to test pypi. So I would say we can go ahead with the merge.

danielhuppmann (approved):

PR #8, opened 2023-08-29
Use `logger.warning` instead of `warn`

danielhuppmann:
I noticed a few deprecation-warnings in the nomenclature nightly tests that ixmp4 uses the deprecated `logger.warn` method instead of `logger.warning`.


danielhuppmann:
Closing in favor of #9

PR #9, opened 2023-08-29
Checking credentials only when needed

danielhuppmann:
Working on the ixmp4-pyam integration, I noticed that ixmp4 immediately (tries to) connect to the IIASA-ScSe-Manager, resulting in a log output like
image

This can be confusing to pyam-only-users, so this PR changes the order so that the credentials are only loaded when needed.


meksor:
Sounds good, that was more the point of the message anyway! I think I might implement a way to use different credentials on the dev manager instance for example, hence the whole 'default'-and-such terminology.

meksor (approved):

PR #11, opened 2023-08-31
Include isort

glatterf42:
### Summary
This PR introduces [isort](https://pycqa.github.io/isort/index.html) as a dependency in the `dev` group. It also tries to update `anyio` and `pandas` to their latest versions allowed by our `pyproject.toml`.

#### Context
isort is a tool for sorting and organizing `import` statements. Code contributions to the MESSAGE stack are checked for compliance, so I think we will want to use it for ixmp4, too, for consistency alone. I also think it does make the `import`s more readable. 
isort offers [compatibility with black](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) and should be easily configurable to run automatically whenever saving code files. For vscode, use the official [isort extension](https://marketplace.visualstudio.com/items?itemName=ms-python.isort) and edit the settings to read
```json
  "[python]": {
    "editor.defaultFormatter": "ms-python.black-formatter",
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
        "source.organizeImports": true
    },
  },
  "black-formatter.importStrategy": "fromEnvironment",
  "isort.args":["--profile", "black"],
  "isort.importStrategy": "fromEnvironment",
```
These settings are partly redundant with our pyproject.toml, which also reads
```pyproject.toml
[tool.isort]
profile = "black"
```
The `importStrategy` ensures that the tools from the virtual environment are used, if they are present. If not, some default versions will be used. `isort.args` enables the black compatibility.

### DIFF
--- a/ixmp4/db/utils/__init__.py
+++ b/ixmp4/db/utils/__init__.py
@@ -1,7 +1,8 @@
 from contextlib import suppress
-from sqlalchemy import sql, inspect
-from sqlalchemy.sql import ColumnCollection
+
+from sqlalchemy import inspect, sql
 from sqlalchemy.orm import Mapper
+from sqlalchemy.sql import ColumnCollection
 
 from ixmp4.core.exceptions import ProgrammingError
 
--- a/ixmp4/__init__.py
+++ b/ixmp4/__init__.py
@@ -1,23 +1,17 @@
 # flake8: noqa
 import importlib.metadata
 
-from ixmp4.core import (
-    Platform as Platform,
-    Region as Region,
-    Unit as Unit,
-    Run as Run,
-    Variable as Variable,
-    Model as Model,
-    Scenario as Scenario,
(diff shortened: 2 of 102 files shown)


ixmp4/db/utils/__init__.py
```diff
@@ -1,7 +1,8 @@
 from contextlib import suppress
-from sqlalchemy import sql, inspect
-from sqlalchemy.sql import ColumnCollection
+
+from sqlalchemy import inspect, sql
 from sqlalchemy.orm import Mapper
+from sqlalchemy.sql import ColumnCollection
```
glatterf42, line 5: Importing both `sql` and `sql.ColumnCollection` explicitly seems a little redundant.

meksor (approved):
Obviously I can't check every single file, but it looks alright from what i can see.

PR #16, opened 2023-09-04
Add Python 3.11 to supported versions

danielhuppmann:
Add Python 3.11 to supported versions


danielhuppmann:
Ok, I have no idea why this PR is failing... Please help @meksor @phackstock @glatterf42!

meksor:
My first guess: we might need the package python3.11-dev in the container used to run these checks.

phackstock:
@meksor, seems like you're onto something https://github.com/aio-libs/aiohttp/issues/6600#issuecomment-1659592087.

phackstock:
Installing python3.11-dev does not seem to be the solution... I'll keep on digging.

phackstock:
@danielhuppmann, tests are finally running now. 
After some painful debugging I was able to fix it.
The issue seems to have been a package called `httptools` which is a dependency of the `uvicorn` package.
`uvicorn` was pinned to an outdated version ^0.15, I now updated it to the closest version that fixed the install which turned out to be 0.17.

phackstock:
Let's see if the tests run through. If they do I guess we can go ahead with the merge, fix the conda publishing and then release v0.3.1.

phackstock (approved):
Approving the PR since the tests are passing.

PR #18, opened 2023-09-08
Drop support for ORACLE

danielhuppmann:
This PR removes the dependency `oracledb` and removes remaining references to ORACLE as a supported database type.


meksor (approved):
lgtm

PR #19, opened 2023-09-14
Fix like lookups not working

meksor:
Like lookups were throwing 500s due to a bug in the filter framework. This PR regresses tests and fixes the bug.


pmussak (approved):
looks good to me

Issue #20, opened 2023-09-15
Install requires building `psycopg2` from source


danielhuppmann:
Per this discussion https://github.com/IAMconsortium/pyam/issues/783 raised by @khaeru, the current dependency on `psycopg2` and `pg_config` is too cumbersome for the wide user base reached via pyam and message-ix.

Therefore, Postgres should be turned into an optional dependency so that users can easily use the Rest-API-client-part of ixmp4.

meksor:
Whoops seems we worked on that at the same time: 
Quoting from the other thread: 

> Actually, a correction on this: psycopg3 IS out, its just not called psycopg3 but reintegrated into the original package. Furthermore, binary installation is no longer discouraged: https://www.psycopg.org/psycopg3/docs/basic/install.html#binary-installation
There is also a pure-python implementation to fall back to but it still requires postgres system dependencies...

PR #21, opened 2023-09-15
Use Psycopg3 instead of Psycopg2

meksor:
Adresses #20 and adjusts the `psycopg2` dependency. 
Using the binary package is no longer discouraged, so this is in use because its easiest and does not require system dependencies. [Read more here.](https://www.psycopg.org/psycopg3/docs/basic/install.html)

Also replaces `postgresql://` platform dsn prefixes with the right ones to keep it compatible with old `platforms.toml` files.


danielhuppmann (commented):
Doesn't a dsn `postgresql+psycopg` make it quite long and cumbersome going forward? Wouldn't it make more sense to use `postgresql+psycopg2` for any "old" database instances and simply use `postgresql` for current ones?

meksor:
That is not our choice, but sqlalchemy's. We could inspect and transform every dsn from our own format so-to-say but i would suspect that is more cumbersome than just using sqlalchemy's way.

danielhuppmann (approved):
Thanks for the quick fix!

PR #25, opened 2023-10-10
Fix potential memory leak

meksor:
Fixes a potential memory leak in the `ManagerConfig` class. 
Since the `lru_cache` decorator is used in the class body the wrapped function and its cache persists across class instances and will not be garbage collected upon deletion of the class instance. 
This PR fixes that by moving the cache construction to the classes `__init__` method.


glatterf42 (approved):
Looks good to me :)

PR #28, opened 2023-11-08
Include optimization scalar

glatterf42:
## Summary

This PR expands the data model required for message_ix by adding limited support for Scalars. One can now create, get, list, tabulate, and manage docs for scalars with the following syntax:

## Implementation Details
```python
# In the core layer:
test_mp = Platform(_backend=RestTestBackend(test_sqlite_mp.backend))
unit = test_mp.units.create("Test Unit")
# To create a new scalar:
# NOTE: the combination of (Run.id, Scalar.name) is checked for uniqueness
        scalar_1 = run.optimization.scalars.create(
            "Scalar 1", value=10, unit="Test Unit"
        )

# To list existing scalars:
scalars = run.optimization.scalars.list()

# To get a single scalar:
scalar = run.optimization.scalars.list(name="Scalar 1")
# or:
scalar = run.optimization.scalars.get("Scalar 1")

# To tabulate a scalar by specifying a name:
scalar_1 = run.optimization.scalars.tabulate(name="Scalar 1")

# To update a scalar:
        scalar = run.optimization.scalars.create(
            "Scalar", value=10, unit=unit.name
        )
        assert scalar.value == 10
        assert scalar.unit.id == unit.id

        scalar.value = 30
        scalar.unit = "Test Unit"

        result = run.optimization.scalars.get("Scalar")

        assert scalar.id == result.id
        assert scalar.name == result.name
        assert scalar.value == result.value == 30
        assert scalar.unit.id == result.unit.id == 1

# See tests/data/test_optimization_scalar.py to see the syntax for the data layer.
```

## Further Changes
While working on this PR, I added/updated a few other minor things:

- Added tests for the data layer functionality of IndexSets
- Updated/expanded IndexSet tests
- Added a data-layer get_by_id function for Units

## TO-DOs

- [x] Add migrations once the DB model is accepted


danielhuppmann:
Thank you @glatterf42!

One misunderstanding that is becoming obvious in this PR - it should not be possible to have several scalars with the same name (and different units).

I imagine the following behavior:

```python
# In the data layer:
test_mp = Platform(_backend=RestTestBackend(test_sqlite_mp.backend))
run = test_mp.backend.runs.create("Model", "Scenario")
unit = test_mp.backend.units.create("Unit")
other_unit = test_mp.backend.units.create("Other Unit")

scalar = run.scalars.create(name="Scalar", value=1, unit_id=unit.id)

# overwrite the scalar
scalar = run.scalars.create(name="Scalar", value=2, unit_id=other_unit.id)

run.scalars.get("Scalar")
> Scalar(name="Scalar", value=2, unit_id=other_unit.id)

# the initial `scalar` object is modified
scalar.value
> 2
```

glatterf42:
Thanks, I updated the functionality as discussed. You would now do/see this:
```python
# In the core layer:
test_mp = Platform(_backend=RestTestBackend(test_sqlite_mp.backend))
run = test_mp.backend.runs.create("Model", "Scenario")
unit = test_mp.backend.units.create("Unit")
other_unit = test_mp.backend.units.create("Other Unit")
scalar = run.optimization.scalars.create(name="Scalar", value=1, unit_id=unit.id)

# overwrite the scalar
scalar = run.optimization.scalars.create(name="Scalar", value=2, unit_id=other_unit.id)
> E           ixmp4.core.exceptions.ScalarNotUnique: Scalar 'Scalar' already exists! Did you mean to call run.optimization.scalars.update()?

scalar = run.optimization.scalars.update("Scalar", value=2, unit_id=other_unit.id)
scalar
> Scalar(name="Scalar", value=2, unit_id=other_unit.id)

# the initial `scalar` object is modified
scalar.value
> 2
```

Note that `run.optimization.Scalar()` still takes a `Unit` to define the scalar, while `run.optimization.scalars.create()` (in the same layer) takes `Unit.id` to define the scalar. I'm still not sure if this is advisable.

The core layer `Scalar` class now only supports `Scalar.unit` as opposed to also `Scalar.unit_id`. This might streamline usage since there is no division between `Scalar.unit_id` (in the core layer) and `Scalar.unit__id` (in the data layers) anymore. However, while all `Scalar` instances should support `Scalar.unit.id`, the core layer does currently not support `Scalar.unit__id` (same for IndexSets, if I remember correctly). Would we want it to?

glatterf42:
Please note that there is support for upsert operations in [sqlalchemy](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-upsert-statements) for [sqlite](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#sqlite-on-conflict-insert) and [postgresql](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#postgresql-insert-on-conflict). I opted for separate `create()` and `update()` functionality because our current `create()` functions employ the `session.add` statement, which seems to be different from the `INSERT` statement that is required for upserts. We would probably also have to manage the different dialect-specific `insert()` functions in order to employ their `insert().on_conflict_do_update()` functionality, since this doesn't exist for all backends. Thus, keeping `create()` and `update()` separate seemed easier and more aligned with current infrastructure. If more update functionality is desired, it might however be beneficial to look into the existing upsert functions and maybe add an `Upserter` mixin.

danielhuppmann:
Thank you, nice!

On the unit-question, if I understand it correctly, being able to pass the unit as a string in the create/update methods would be much more user-friendly...

One question on the preferred API (maybe to be discussed with @meksor):
Do we want
```python
Run.scalars.update("Scalar",value=3, unit="some unit")
```
or 
```python
Run.scalars.get("Scalar").update(value=3, unit="some unit")
```
or 
```python
Run.scalars.get("Scalar").value = 3
Run.scalars.get("Scalar").unit = "some unit"
```
?

glatterf42:
Done, assuming you mean that
```python
Run.scalars.get("Scalar").value = 3
Run.scalars.get("Scalar").unit = "some unit"
```
should not only change the local version of the object, but also the one stored in the DB. So now, test of this form pass:
```python
        run = test_mp.Run("Model", "Scenario", "new")
        unit = test_mp.units.create("Test Unit")
        unit2 = test_mp.units.create("Test Unit 2")
        scalar = run.optimization.scalars.create(
            "Scalar", value=10, unit_name=unit.name
        )
        assert scalar.value == 10
        assert scalar.unit.id == unit.id

        scalar = run.optimization.scalars.update(
            "Scalar", value=20, unit_name=unit2.name
        )
        assert scalar.value == 20
        assert scalar.unit.id == unit2.id

        scalar.value = 30
        scalar.unit = "Test Unit"
        # NOTE: doesn't work for some reason (but doesn't either for e.g. model.get())
        # assert scalar == run.optimization.scalars.get("Scalar")
        result = run.optimization.scalars.get("Scalar")
        assert scalar.id == result.id
        assert scalar.name == result.name
        assert scalar.value == result.value
        assert scalar.unit.id == result.unit.id
```

#### Further implementation updates

We no longer need to call `self.backend.units.get_by_id()` in the Scalar core layer :)

ixmp4/db/__init__.py
```diff
@@ -49,6 +49,7 @@
     update,
 )
 from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.exc import MultipleResultsFound
```
meksor, line 52: Can't seem to find this used anywhere? (ctrl+f on the review page)

ixmp4/server/rest/optimization/scalar.py
```diff
@@ -0,0 +1,75 @@
...
+    ),
+    table: bool = Query(False),
+    backend: Backend = Depends(deps.get_backend),
+):
+    return backend.optimization.scalars.enumerate(_filter=filter, table=bool(table))
+
+
+@autodoc
+@router.patch("/{id}/", response_model=api.Scalar)
+def update(
```
meksor, line 62: This endpoint requires the `id`, `run__id` and `name`? Isn't that kind of redundant?
Maybe its worth refactoring the update method to only take the `id` instead of `name` and `run__id`? 
Would make some sense since right now its hard to distinguish between arguments that are going to be updated and the ones used to fetch the actual instance.

tests/core/test_indexset.py
```diff
@@ -1,47 +1,105 @@
...
+    )
+
+
 @all_platforms
 class TestCoreIndexSet:
     def test_create_indexset(self, test_mp):
         run = test_mp.Run("Model", "Scenario", "new")
-        index_set_1 = run.optimization.IndexSet("IndexSet 1")
-        returned_index_set_1 = run.optimization.IndexSet("IndexSet 1")
-        assert index_set_1.id == returned_index_set_1.id
```
meksor, line 12: Why did you remove this test case?

tests/core/test_indexset.py
```diff
@@ -1,47 +1,105 @@
...
+
+    def test_tabulate_indexsets(self, test_mp):
+        run = test_mp.Run("Model", "Scenario", "new")
+        # Per default, tabulate() lists only `default` version runs:
+        run.set_as_default()
+        indexset_1 = run.optimization.IndexSet("Indexset 1")
+        indexset_2 = run.optimization.IndexSet("Indexset 2")
+        exp = df_from_list(indexsets=[indexset_1, indexset_2])
+        res = run.optimization.indexsets.tabulate()
+        # utils.assert_unordered_equality doesn't like lists, so make sure the order in
```
meksor, line 93: What does "doesnt like lists" mean? assert_unordered_equality is meant only for DFs afaik...

ixmp4/core/optimization/scalar.py
```diff
@@ -0,0 +1,172 @@
...
+    _unit: Unit
+    NotFound: ClassVar = ScalarModel.NotFound
+    NotUnique: ClassVar = ScalarModel.NotUnique
+
+    def __init__(
+        self,
+        name: str,
+        _run: Run,
+        value: float | None = None,
+        unit_name: str | None = None,
```
meksor, line 24: Wondering if it might make sense to take a unit object here directly instead of just the name?
Effectively while loading data you will have the unit model somewhere already anyway...

tests/data/test_optimization_indexset.py
```diff
@@ -0,0 +1,92 @@
+import pytest
+
+from ixmp4 import IndexSet
+
+from ..utils import all_platforms
+
+
+@all_platforms
+class TestDataOptimizationIndexSet:
```
meksor, line 9: Tabulation tests missing?

tests/data/test_optimization_scalar.py
```diff
@@ -0,0 +1,66 @@
+import pytest
+
+from ixmp4 import Scalar
+
+from ..utils import all_platforms
+
+
+@all_platforms
+class TestDataOptimizationScalar:
```
meksor, line 9: Ditto. Tabulation tests missing?

meksor (changes requested):

ixmp4/core/optimization/scalar.py
```diff
@@ -0,0 +1,172 @@
...
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        self._run = _run
+        if getattr(self, "_model", None) is None:
+            if (value is None) or (unit_name is None):
+                raise TypeError("`Scalar` requires `value` and `unit_name`!")
+            try:
+                self._model = self.backend.optimization.scalars.get(
```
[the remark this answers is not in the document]
danielhuppmann: Not a hard error, we want to allow users to change scalars. But a warning could be useful when the unit changes...

tests/core/test_indexset.py
glatterf42: Line 10 is still in there, just as line 38 now. The other two lines I removed because they are not testing the creation of IndexSets more, but getting one from the DB, which is covered in other tests. For this test to be purely about creation, I removed the `get` parts. This distinction became clear to me while writing the Scalar tests, so I went back and updated the IndexSet tests as well.

tests/core/test_indexset.py
glatterf42: It is, that's why it didn't work when I tried to pass in a list of DFs. That's why I added this auxiliary `df_from_list()` function. We could rephrase this hint as "... only accepts pd.DataFrames,". Or we remove this comment entirely, but both the order of `indexset_1, indexset_2` as arguments for `df_from_list()` as well as the order of `indexset.name, indexset.elements, ...` in the function's return-value-DF are important for the test to work.

ixmp4/core/optimization/scalar.py
glatterf42: I was wondering the same thing, but @danielhuppmann said three weeks ago:
> On the unit-question, if I understand it correctly, being able to pass the unit as a string in the create/update methods would be much more user-friendly...

And I figure, if you have the `unit` object, it's not that much of a difference whether you pass in `unit` itself or `unit.name`.

ixmp4/data/db/run/model.py
```diff
@@ -39,6 +40,7 @@ class Run(base.BaseModel):
     )
 
     indexset: types.Mapped[list["IndexSet"]] = db.relationship()
+    scalar: types.Mapped[list["Scalar"]] = db.relationship()
```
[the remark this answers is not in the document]
glatterf42: Yes, thanks, good catch.

ixmp4/server/rest/optimization/scalar.py
glatterf42: It almost is, but I'm not sure how we can change it. In the core layer, we receive `run.optimization.scalars.update(scalarname, value, unitname)`. Before calling the data layers, we add `run.id` to the call because we have that available for free. 
In the DB layer, we then already have everything we need to update the scalar as desired. But in the API layer, we currently need to call `super.get()` before going to the server/rest layer since the scalars are stored under `/optimization/scalar/{id}`, so we need to provide the ID as well to target the correct address. We can't really use the more general `/optimization/scalar/` address, I think, because the `patch` method for that is already in use for `query()`. 
So if we want to keep the facade syntax with `scalarname, value, unitname` (which seems convenient) and don't want to call the data layer twice (which is not necessary for direct contact with the DB layer), I don't immediately see how the refactoring would look like. However, maybe one of these conditions is not worth keeping, especially the latter? Would you prefer to `get()` a scalar in the core layer from the data layer so that we can provide its `id` to the `update()` function (which would then be enough to replace `name` and `run_id` in all `update()` calls)?

tests/data/test_optimization_indexset.py
glatterf42: Thanks, I included them now.

meksor:
I've added some changes to `ScalarRepository.update`, feel free to revert the changes as you see fit, but i think this way is a little more consistent. It now takes the scalar id as the only positional argument and the rest as kwargs (the values to be updated). 
This means now we actually use the id in the rest endpoint and there is a clear seperation between values to be updated and values use to fetch the instance. 
I've also removed the use of the sqlalchemy `.returning` feature as it does not work with my version of sqlite.
Since i found it confusing that `scalar.unit = scalar.unit` would throw an error I added the ability to directly set a unit object. 

I think its worth reconsidering the creation logic a bit. Right now, if i call `Run.scalars.create` ixmp4 will fetch the scalar and then pass it to the scalar class which in turn fetches the same scalar again. Needs some reworking... 
We should also decide on ONE way to do things like that. EITHER we use a constructor to create/retrieve scalars OR we use the Repository methods. Both seems to lead to confusion and bad code... 
I'd probably opt for the repository method for IndexSet and Scalar since the constructor method is now only used for `Run` which is a special class that manages many other things as well. IndexSet and Scalar dont.

meksor:
(test are failing because the manager is down)

ixmp4/core/optimization/scalar.py
danielhuppmann: When adding an IAMC datapoint, the dataframe must have a column "unit" which is actually the string. I'd like to follow that easy-to-use approach here and have a required argument "unit" which can take a Unit object or a name as string or None for dimensionless.

glatterf42: This is done, though I'll leave a note that we should provide logging information to users forgetting to set `unit_or_name` that we will assume the Scalar to be dimensionless.

glatterf42:
@danielhuppmann: Trying to handle the dimensionless units as you suggested, I've run into the following question: When we call `run.units.create("")`, this creates a dimensionless `Unit` which can later be used by our code. I'm not sure if it's intuitive to require creating a `Unit` for things that are dimensionless. So for now, you can call 
```python
unit = run.units.create("unit")

scalar_1 = run.optimization.scalars.create("Scalar 1", value=1, unit_or_name=unit)
scalar_2 = run.optimization.scalars.create("Scalar 2", value=2, unit_or_name=unit.name)
scalar_3 = run.optimization.scalars.create("Scalar 3", value=3)
```
because the core layer notices that `unit_or_name` is missing, assumes a default `None`, creates a `Unit` called "dimensionless" for you and adds it to the `Scalar`.
However, as @meksor correctly pointed out, we don't do this for any other unit and would have to provide logging information about this at least (people might also just not know/forget to provide a unit). 

Would you like to keep the behaviour as it is or should we require the creation of a "dimensionless" unit before it can be used in a `Scalar`?

glatterf42:
@danielhuppmann I've updated this PR, so the syntax at the top is up to date. There is one question remaining, the other open things are a Todo for myself and a note for the future. Once the question is clarified, I can request meksor to approve this :)

danielhuppmann:
Thanks @glatterf42 for rebasing and updating the PR. I think it's fine either way, whether the dimensionless unit comes pre-defined in a blank database or hast to be defined by a user. Whatever you prefer.

But I still think that the keyword argument should be `unit` instead of `unit_or_name` and it should take a `ixmp4.Unit` or a string. I believe that an argument `unit_or_name` creates more confusion than an argument `unit` that can handle multiple "forms" of a unit object.

glatterf42:
Thanks, then I'll keep the dimensionless unit as pre-defined. The syntax changed now so that you can supply `Scalar`s with `unit`, which can either be a str or a `Unit` object. 

@meksor, please update your review as all your points should be addressed by now. The last thing from my side is adding another DB migration.

glatterf42:
In fact, I don't think we need another migration here. There were some commits after https://github.com/iiasa/ixmp4/pull/28/commits/d66a0a4286faaa9873073f721918793f7bfba5e3 that affected `data/db/optimization/scalar/model.py`, but none that affected it in a way that changed the actual data model. When I ran `alembic revision -m "Update optimization_scalar table" --autogenerate` right now, it created a migration with an almost empty file, like 
```python
def upgrade():
    pass

def downgrade():
    pass
```
So I figure we don't need to push that migration here.

glatterf42:
Rebased onto latest main.

meksor (approved):
Looks good thanks! Sorry for the delay!

PR #30, opened 2023-11-15
Quickfix for numpy.types as meta indicators

danielhuppmann:
This PR implements a quick-fix when setting meta-indicators with numpy-types, as are used in pd.DataFrames. The PR also improves handling of `None`, so that these values are ignored when setting from a dictionary `run.meta = {"key": None}` and are used as "delete" when using the setter like `run.meta["key"] = None`.

### DIFF
--- a/ixmp4/core/run.py
+++ b/ixmp4/core/run.py
@@ -131,7 +131,9 @@ class RunMetaFacade(BaseFacade, UserDict):
         except KeyError:
             pass
 
-        self.backend.meta.create(self.run.id, key, numpy_to_pytype(value))
+        value = numpy_to_pytype(value)
+        if value is not None:
+            self.backend.meta.create(self.run.id, key, value)
         self.df, self.data = self._get()
 
     def __delitem__(self, key):
@@ -142,4 +144,9 @@ class RunMetaFacade(BaseFacade, UserDict):
 
 def numpy_to_pytype(value):
     """Cast numpy-types to basic Python types"""
-    return value.item() if isinstance(value, np.generic) else value
+    if value is np.nan:  # np.nan is cast to 'float', not None
+        return None
+    elif isinstance(value, np.generic):
+        return value.item()
+    else:
+        return value
--- a/ixmp4/core/run.py
+++ b/ixmp4/core/run.py
@@ -112,7 +112,6 @@ class RunMetaFacade(BaseFacade, UserDict):
         return df, dict(zip(df["key"], df["value"]))
 
     def _set(self, meta: dict):
-        print("You are here")
         df = pd.DataFrame({"key": self.data.keys()})
         df["run__id"] = self.run.id
         self.backend.meta.bulk_delete(df)
@@ -125,7 +124,6 @@ class RunMetaFacade(BaseFacade, UserDict):
         self.df, self.data = self._get()
 
     def __setitem__(self, key, value: int | float | str | bool):
-        print("You are there")
         try:
             del self[key]
         except KeyError:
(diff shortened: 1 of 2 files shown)


glatterf42:
The tests all seem to pass in the GHA, so what exactly is not working with `Run.meta[<key>] = value`? Also, what do you want to do and is there no way to use numpy-types for it?

danielhuppmann:
The tests only pass because of the (very not elegant) casting in __setitem__. And I’m sure they would fail if we added numpy.float to the tests…

Proposal: anyone recommends a more pythonic approach to the casting (or something else), then I implement it and add tests for other numpy types.

glatterf42:
[This question](https://stackoverflow.com/questions/9452775/converting-numpy-dtypes-to-native-python-types) seems to be very relevant here.
One option would be to have a map, like you say, but if that map is not created automatically, it might be tedious to cover all cases (because you'd probably want to catch `np.float32` and `np.float64` and some others as `float`). 
Another suggested answer is something like
```python
if isinstance(value, np.generic): 
    value = value.item()           
```
This approach should catch all simple data types, but might not work for `np.ndarrays` or so. Also, some other answers claim it's not the fastest way.
So yet another approach might be
```python
if isinstance(value, np.generic): 
    value = value.tolist()  
```

danielhuppmann:
Thanks @glatterf42, followed one of your suggestions... 

I also added handling of `np.nan`/`None` and  `Run.meta["<some key>"] = None`  can now be used to delete an entry.

For explanation, having a dictionary with `None`-values may seem non-sensical, but it is a real issue when you have a **pyam.IamDataFrame** where one scenario (=run) does not have values for some indicators. See https://github.com/IAMconsortium/pyam/pull/797 for where this is relevant.

ixmp4/core/run.py
```diff
@@ -111,24 +112,41 @@ def _get(self) -> tuple[pd.DataFrame, dict]:
         return df, dict(zip(df["key"], df["value"]))
 
     def _set(self, meta: dict):
+        print("You are here")
```
glatterf42, line 115: This line should be deleted again:
```suggestion
```

ixmp4/core/run.py
```diff
@@ -111,24 +112,41 @@ def _get(self) -> tuple[pd.DataFrame, dict]:
...
+        df = pd.DataFrame(
+            {"key": meta.keys(), "value": [numpy_to_pytype(v) for v in meta.values()]}
+        )
+        df.dropna(axis=0, inplace=True)
         df["run__id"] = self.run.id
         self.backend.meta.bulk_upsert(df)
         self.df, self.data = self._get()
 
     def __setitem__(self, key, value: int | float | str | bool):
+        print("You are there")
```
glatterf42, line 128:
```suggestion
```
As should be this one.

glatterf42 (commented):
Apart from the requested deletions, this looks good to me. Just one clarifying question: is it enough to test `np.float64` or should we investigate why `np.float128` is not working and fix that?

danielhuppmann:
Sorry, removed the print statements.

About numpy.float128, this is converted to a longdouble, wich then creates issues in the database layer. And numpy.float32 is not "precise", writing 13 to the db and reading it again returns 12.9...

Given that pandas uses float64 internally as a default, this PR is good enough for proceeding with the pyam integration. We can tackle more cases once it's necessary.

glatterf42 (approved):
Thanks for the explanation, looks good to me, then :)

meksor (approved):
lgtm

PR #31, opened 2023-11-23
Change filter-wildcard to `*`

danielhuppmann:
The pyam package and the manager-service use `*` as a wildcard, whereas SQL uses `%`.

This PR changes the API to be consistent with pyam, and also adding escaping to the string-like-filters so that it is possible to filter (only) for the percentage sign as a unit.


meksor (approved):
lgtm

PR #34, opened 2023-12-11
Implement `map()` method for model and scenario repositories

danielhuppmann:
This PR implements (only) the `map()` methods from #33.

### DIFF
--- a/ixmp4/core/model.py
+++ b/ixmp4/core/model.py
@@ -74,6 +74,9 @@ class ModelRepository(BaseFacade):
     def tabulate(self, name: str | None = None) -> pd.DataFrame:
         return self.backend.models.tabulate(name=name)
 
+    def map(self) -> dict:
+        return self.backend.scenarios.map()
+
     def _get_model_id(self, model: str) -> int | None:
         if model is None:
             return None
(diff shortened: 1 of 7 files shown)


ixmp4/core/model.py
```diff
@@ -74,6 +74,9 @@ def list(self, name: str | None = None) -> Iterable[Model]:
     def tabulate(self, name: str | None = None) -> pd.DataFrame:
         return self.backend.models.tabulate(name=name)
 
+    def map(self) -> dict:
+        return self.backend.scenarios.map()
```
meksor, line 78: Is it necessary to expose map to the facade?
Also: typo: should be .models.map(), right?

ixmp4/core/scenario.py
```diff
@@ -74,6 +74,9 @@ def list(self, name: str | None = None) -> Iterable[Scenario]:
     def tabulate(self, name: str | None = None) -> pd.DataFrame:
         return self.backend.scenarios.tabulate(name=name)
 
+    def map(self) -> dict:
+        return self.backend.scenarios.map()
```
meksor, line 78: ditto

meksor (changes requested):

ixmp4/core/model.py
meksor: On that note: there should probably be some tests for these.

danielhuppmann: Removed from the facade.

ixmp4/core/scenario.py
danielhuppmann: Removed from the facade

danielhuppmann:
Removed the `map()` from the facade, added tests for the core

meksor (approved):
Lgtm

PR #35, opened 2023-12-12
Refactor the facade API to create/get runs

danielhuppmann:
As just discussed, this PR changes the API to create/get runs from the (implicit)

```python
Platform.Run(model, scenario, version="new")
```
to the more pythonic 
```python
Platform.runs.create(model, scenario)
```
and a corresponding get
```python
Platform.runs.get(model, scenario)
```

Maybe I'm a bit old-fashioned, but I'd still like to keep the default vs. non-default selection implicit in the `get()` method on the facade-layer.

fyi @pmussak @glatterf42

### DIFF
--- a/ixmp4/core/run.py
+++ b/ixmp4/core/run.py
@@ -65,7 +65,7 @@ class RunRepository(BaseFacade):
         self,
         model: str,
         scenario: str,
-    ):
+    ) -> Run:
         return Run(
             _backend=self.backend, _model=self.backend.runs.create(model, scenario)
         )
@@ -75,7 +75,7 @@ class RunRepository(BaseFacade):
         model: str,
         scenario: str,
         version: int | None = None,
-    ):
+    ) -> Run:
         if version is None:
             _model = self.backend.runs.get_default_version(model, scenario)
         else:
--- a/ixmp4/core/run.py
+++ b/ixmp4/core/run.py
@@ -18,29 +18,10 @@ class Run(BaseModelFacade):
     NotFound: ClassVar = RunModel.NotFound
     NotUnique: ClassVar = RunModel.NotUnique
 
-    def __init__(
-        self,
-        model: str | None = None,
-        scenario: str | None = None,
-        version: int | str | None = None,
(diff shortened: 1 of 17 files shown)


tests/core/test_run.py
```diff
@@ -22,14 +22,14 @@ def _expected_runs_table(*row_default):
 @all_platforms
 class TestCoreRun:
     def test_run_versions(self, test_mp):
-        run1 = test_mp.Run("Model", "Scenario", version="new")
```
meksor, line 25: We should probably add a test for the `Run.NotFound` exception here now?

meksor (changes requested):

tests/core/test_run.py
danielhuppmann: Thanks, added a test

ixmp4/core/run.py
```diff
@@ -80,6 +61,27 @@ def unset_as_default(self):
 
 
 class RunRepository(BaseFacade):
+    def create(
+        self,
+        model: str,
+        scenario: str,
+    ):
```
glatterf42, line 68: We can type hint this as 
```suggestion
    ) -> Run:
```

ixmp4/core/run.py
```diff
@@ -80,6 +61,27 @@ def unset_as_default(self):
...
+        return Run(
+            _backend=self.backend, _model=self.backend.runs.create(model, scenario)
+        )
+
+    def get(
+        self,
+        model: str,
+        scenario: str,
+        version: int | None = None,
+    ):
```
glatterf42, line 78: Same here:
```suggestion
    ) -> Run:
```

glatterf42 (approved):
Thanks, looks good to me. One note, though: at the moment, the only place where we test providing the `version` parameter to `runs.get()` is in `test_benchmarks`, so it might seem untested by simply looking at `test_run`. It _is_ tested, so things are fine, but the test location is not very intuitive.

meksor (approved):
LGTM

PR #37, opened 2024-01-04
Show extended run-"index" from `Platform.meta.tabulate()`

danielhuppmann:
Per review comment by @meksor at https://github.com/iiasa/ixmp4/pull/33#issuecomment-1829810151, this PR has an alternative implementation to show the model-scenario-version "index" of the meta dataframe.

The joining of columns is now done on the database layer using a new argument `join_run_index`, following the logic that the model-scenario[-version] columns are called "index" in pyam. [Alternative suggestions for the argument-name welcome.]

The argument is optional on the backend, defaulting to show only the run-id. But when calling the meta-dataframe from the facade, the extended "index" is returned.


meksor (approved):
lgtm

PR #38, opened 2024-01-09
Add a `raw` argument to the `iamc.tabulate()`

danielhuppmann:
As discussed today, this PR adds an argument `raw` to the `iamc.tabulate()` endpoint - alternative suggestions for the name welcome...

At the end of a series PRs, the iamc endpoints should take (`add()`, `remove()`) and return (`tabulate()`) both the ixmp4-db-internal format (with columns type, step_year, step_datetime, step_category) or the standard IAMC format with columns year, time and/or subannual.

This PR cleans up the datapoints-tets and adds the feature to correctly return the datapoints in standard IAMC format. In the interest of keeping this PR manageable, I did not add more changes.


meksor (commented):
Hm, not sure about making the default different, my guess is that the dependants will rely on the current format so it is probably better to make raw=True the default or rename the argument to `iamc_format` and default to false.
Also, there seem to be misordered imports...

danielhuppmann:
I'm a bit skeptical of having an argument `iamc_format` within the `iamc` namespace - seems confusing to me.

Also, I don't think that there are any dependants yet anyway at the facade level, pyam will be the first. And I see a strong case for making the default such that it is like the format that people are used to, both for input to and return from ixmp4.

meksor (approved):
Oh, right sorry, I was confused and thought this was in the db layer. 
LGTM in that case!

PR #40, opened 2024-01-26
Include optimization.table

glatterf42:
The next part of adding the message_ix/ixmp data model. 

Still to be done:

- [x] Add list, enumerate, tabulate, docs functionality in DB layer
- [x] Add API layer
- [x] Add Core layer

For now, still a WIP.


glatterf42:
While this PR is still a WIP, you can already take a look at what is arguably the most interesting part: adding data in the DB layer to a Table, which requires lots of validation. The following behaviour and possibly more can be inferred from the existing tests and files in the data/abstract and data/db directories.

<details>

<summary>Current details on adding and validating data</summary>

### How does it work?

Currently, data can be provided as either a dict or a `pd.DataFrame`. Note that when creating a dict with values of `len() == 1`, you still need to provide them as lists, like `dict = {"key": [value]}`. This is because we use pandas and its existing functionality for validation, for which we need to convert the dict to a `DataFrame`, which requires a fixed order. 

Before validation, though, we use the dict form of the data to merge them with possibly pre-existing Table.data. This is done via the union operator `dict1 | dict2` for dicts, which overwrites keys in `dict1` if `dict2` also provides them, so this already goes in the direction of updating data and probably warrants logging information at least.

Whenever table.data is set (not that `table.data |= data` does not count as setting, while `table.data = table.data | data` does; similar to adding `elements` to `IndexSet`s), the `validate_data()` function is triggered and checks that 

- no values are missing (each column contains the same number of values);
- there are no duplicate rows in the data;
- all columns only contain data allowed as per the linked `Indexset.elements`

At the moment, all of these cases raise `ValueError`s (though with differing messages); I don't know if we want to raise custom errors instead.

`IndexSet`s are constrained when the Table is created. Per default, the `IndexSet.name`s become the column names, but this can be overwritten when specifying `column_names=[...]`. In that case, `column_names` needs to contain exactly one unique name for each column.

#### Some more edge case considerations
##### `Data.keys != len(constrained_to_indexsets)`
I've tried to ensure that several edge cases are covered already. For example, if you provide a data dict that doesn't contain data for all specified columns, nothing happens. We should introduce a check that all required data is present before the Table is used elsewhere, but it's fine to add the data piecemeal. If you provide duplicate keys in the same dict, the latter will overwrite the former, but tools like ruff will warn you before that happens.

###### Using an `IndexSet` twice in `constrained_to_indexsets`
For now, this is fine. I think something like a `model_years` indexset might be used for something like `year_active` and `year_vintage`, but @danielhuppmann, please let me know if this should be constrained as well.

##### Duplicates in `column_names`
Will raise a `ValueError`. 

##### Different data types
As one test demonstrates (with `table_5`, I think), a column can contain data of different types. At one point, I noticed `Column.dtype` being a strange `object`, but I've not studied this further for now since I don't know what we'll be using the `dtype` for and it's not currently in use, anyway. @meksor, maybe you could tell me again what the `dtype` and `unique` parameters from the [requirements](https://wiki.ece.iiasa.ac.at/wiki/index.php/IXMP4/MESSAGEix-Requirements) were supposed to achieve? I might be missing more than I think.

</details>

glatterf42:
## Questions and possible ToDos
@meksor and @danielhuppmann, sorry for the long post. While cleaning up this PR, I came across the following questions/notes I took during its implementation. I'd like to clarify them before merging the PR even if our solution is to open a new issue and include the fix/expansion in another PR.

#### Adding data to a Table in the data layer
We currently have the following syntax for that (very similar to how elements are added to an Indexset):
```python
        table = test_mp.backend.optimization.tables.create(
            run_id=run.id,
            name="Table",
            constrained_to_indexsets=[indexset_1.name, indexset_2.name],
        )
        test_mp.backend.optimization.tables.add_data(
            table_id=table.id, data=test_data_1
        )
        table = test_mp.backend.optimization.tables.get(run_id=run.id, name="Table")
        assert table.data == test_data_1
```
I'm wondering if we want to keep it that way. In the core layer, we have
```python
        table = run.optimization.tables.create(
            "Table",
            constrained_to_indexsets=[indexset.name, indexset_2.name],
        )
        table.add(data=test_data_1)
        assert table.data == test_data_1
```
Note that this only works because `table.add()` both adds the data and calls `get()` to return the updated Table.

#### Linking each to Column to a unique Indexset
I'm wondering if different Columns of a Table can be constrained to the same Indexset. Or in other words, I'm wondering if the following should raise an error:
```python
        with pytest.raises(ValueError):
            _ = test_mp.backend.optimization.tables.create(
                run_id=run.id,
                name="Table 2",
                constrained_to_indexsets=[indexset_1.name, indexset_1.name],
                column_names=["Column 1", "Column 2"]
            )
```

#### Raising distinct error messages when validating data
At the moment, we have numerous validation checks for data that is being added to a Table, but all of these checks only raise `ValueError`s (though with unique messages):
https://github.com/iiasa/ixmp4/blob/b682f7e04a860b797135f9ba083da3bfa6685bef/ixmp4/data/db/optimization/table/model.py#L48-L71
And
https://github.com/iiasa/ixmp4/blob/b682f7e04a860b797135f9ba083da3bfa6685bef/ixmp4/data/db/optimization/table/repository.py#L79-L97
So my question is: are we fine with that or would we want to have distinct custom errors for all these checks?


#### Allowing Table.data to be added piece-meal 
Currently, we allow to add data for the various Columns piece-meal like so:
```python
        table_3 = run.optimization.tables.create(
            name="Table 3",
            constrained_to_indexsets=[indexset.name, indexset_2.name],
            column_names=["Column 1", "Column 2"],
        )
        table_3.add(data={"Column 1": ["bar"]})
        assert table_3.data == {"Column 1": ["bar"]}

        table_3.add(data={"Column 2": [2]})
        assert table_3.data == {"Column 1": ["bar"], "Column 2": [2]}

        table_3.add(
            data=pd.DataFrame({"Column 1": ["foo"], "Column 2": [3]}),
        )
        assert table_3.data == {"Column 1": ["foo"], "Column 2": [3]}
```
If the data to be added contains data for a Column that already has data, that Column's data is overwritten. 
If we are fine with this behavior, we should probably still

- [ ] Add functionality to check which data is already present/still needed (we might also want to somehow include information on how to see which data are permitted by an Indexset)
- [ ] Log information if a Column's data is overwritten
- [ ] Add a check that all data is present before the Table is used elsewhere (though I'm not sure this is possible since there's no clear endpoint of "this is where we stop adding data to the Table", so it's not clear to me when this kind of test should occur)


#### Specifying the type of `constrained_to_indexsets`
In the core layer, we currently support `tables.create()` to receive the parameter `constrained_to_indexsets` as a list of strings:
```python
def create(
        self,
        name: str,
        constrained_to_indexsets: list[str],
        column_names: list[str] | None = None,
    ) -> Table:
```
However, we return the property `Table.constrained_to_indexsets` as a list of integers:
```python
    def constrained_to_indexsets(self) -> list[int]:
        return [column.constrained_to_indexset for column in self._model.columns]
```
Which is the preferred form? Or should we accept `list[str | int]` for `create()` and still only return `int`?
In the DB layer, such a property does not exist. A `Table` has `Columns` and each `Column` has `constrained_to_indexset`, which is of type `int`. I only added the property in the core layer on a whim because I thought it might be convenient.

glatterf42:
#### Handling Columns of Tables
At the moment, Columns are only added to a Table during the creation of the Table. Do we want to keep it that way?
We could alternatively think about allowing Columns to be added to Tables manually after their creation and also then allow removing Columns from Tables (which might entail removing data).

danielhuppmann:
Thanks @glatterf42, see my responses below:

## Adding data to a Table in the data layer

Yes, this approach makes sense to me.

You could improve performance by setting `table` to None after an `add()`, and only `get()` when `table` is explicitly accessed afterwards. This way, a user wouldn't have to load the entire table multiple times.

```python
def add():
    test_mp.backend.optimization.tables.add_data(
        table_id=table.id, data=test_data_1
    )
    self._table = None
```

@getter 
def table:
    if self._table is None:
        self._table = test_mp.backend.optimization.tables.get(run_id=run.id, name="Table")
     return self_table
````

## Linking each to Column to a unique Indexset

Yes, it is absolutely crucial that multiple columns can be foreign-keyed to the same index set. Only the name has to be unique.

## Raising distinct error messages when validating data

Good enough for now, but maybe to be improved later.

## Allowing Table.data to be added piece-meal

Fine in principle to add data step by step similar to the current behavior, but data should be added row-wise. So if you have a table with two columns, the following line should raise an error.

```python
table_3.add(data={"Column 1": ["bar"]})
assert table_3.data == {"Column 1": ["bar"]}
> ValueError Missing entry for 'Column 2'
```

## Specifying the type of constrained_to_indexsets

I assume that the `int` is the unique id of an indexset? In terms of usability, I would only show the (unique) name of an indexset to a user.

## Handling Columns of Tables

It should not be possible to add columns to a table after creation.

glatterf42:
Thanks for the answers @danielhuppmann :)

#### Adding data to a Table in the core layer
Thanks for the suggested performance improvement. I didn't implement it yet because the `add()` function is a member of the Table class and I don't think we want to have a `Table.table` attribute. The current workflow would be:
```python
        test_mp: Platform = request.getfixturevalue(test_mp)  # type: ignore
        indexset = run.optimization.indexsets.create("Indexset")
        indexset.add(elements=["foo", "bar"])
        indexset_2 = run.optimization.indexsets.create("Indexset 2")
        indexset_2.add([1, 2, 3])
        test_data_1 = {"Indexset": ["foo"], "Indexset 2": [1]}
        table = run.optimization.tables.create(
            "Table",
            constrained_to_indexsets=[indexset.name, indexset_2.name],
        )
        table.add(data=test_data_1)
        assert table.data == test_data_1
```
I'm not sure if we could hold off the data loading until `table.data` is actually being used, but I agree we might be able to save some memory usage here in theory. 

#### Allowing Table.data to be added piece-meal
There is another validation check now that catches data-setting attempts where data doesn't have enough keys to serve all Table.Columns: https://github.com/iiasa/ixmp4/blob/922f9588952c48cf4010e0d40bae6d9a15e2923f/ixmp4/data/db/optimization/table/model.py#L48-L55

#### Specifying the type of constrained_to_indexsets
We now return the names for convenience :)

#### Handling Columns of Tables
`backend.optimization.tables.add_column()` has been renamed to `backend.optimization.tables._add_column()`, indicating the intent that this is a private function. I've also removed the API- and Docs-layer mentions of this function and instead included a shorter version of its docstring [in the DB layer](https://github.com/iiasa/ixmp4/blob/922f9588952c48cf4010e0d40bae6d9a15e2923f/ixmp4/data/db/optimization/table/repository.py#L31-L58). Hope this is okay @meksor.

glatterf42:
To be clear about the current workflow to add data to a table: you'd create the table and then need to call `add_data()` with _all the data_ you want to add. If you add something like `{"test": 2, "nexttest": 3}` and then decide you want to use also/instead add `{"test": 4, "nexttest": 5}`, we currently take that as an "instead" and overwrite the existing keys. So if you wanted to truly add to the existing data, you'd need to add `{"test": [2, 4], "nexttest": [3, 5]}`. 
I don't know if such a workflow is realistic or if people should be able to expect to add to any existing data, maybe @danielhuppmann can clarify this and maybe @meksor has an opinion on which expectation people should have.

Either way, https://github.com/iiasa/ixmp4/blob/ee3d437bc865a1fa983373572fb83750b62ade52/ixmp4/data/db/optimization/table/repository.py#L154 seems kind of unnecessary now. I introduced this back when I thought it should be able to add pieces of one row and then add the remaining columns later on. However, for overwriting, we can also simply write `table.data = data`. If we want to truly add data row-wise, I might need to come up with something a little more fancy. The question then is whether this should be part of this PR; with ideal trunk-based development the answer is probably no, but this PR has already deviated from that anyway and so I might as well "finish" the feature here (I'm well aware that you can't remove any optimization data from the DBs at the moment).

But otherwise, I think this PR is ready for review :)

danielhuppmann:
Thanks @glatterf42 

> To be clear about the current workflow to add data to a table: you'd create the table and then need to call add_data() with all the data you want to add. If you add something like {"test": 2, "nexttest": 3} and then decide you want to use also/instead add {"test": 4, "nexttest": 5}, we currently take that as an "instead" and overwrite the existing keys. 

No, this is not the expected use case, given current behavior of ixmp and usual modelling workflows.

Imagine a "table" being the investment costs for different power plants in several regions. You often have situations where a modeler has a script e.g. computing parameters for only one type of power plant (or one region).

Similar to the current MESSAGE tutorials, I see the need for a workflow like

```python
inv_cost_wind = pd.DataFrame()  # get a dataframe of investment cost parameters for a certain technology
run.parameters.get("inv_cost").add_data(inv_cost_wind)
```
adding (or replacing  existing) datapoints but not removing existing parameter datapoints.

glatterf42:
Sorry, I don't quite follow: in your example, would `inv_cost_wind` contain all the data currently in the parameter plus the updated values or would it only contain the values that need updating?

And mainly for me, to clarify: I'm imagining a table like this:
| power plant type  | Region | Investment cost |
| ------------- | ------------- | -------------|
| wind  | EEU  | x |
| wind | WEU | y |
| water  | EEU | z |

And a modeler might now want to update `wind` in `EEU` to `w`, so they would only provide the `w` value (plus other columns required for this one row) and expect this to overwrite the existing value for that row, while leaving the rest untouched?

danielhuppmann:
The usual workflow is to only provide the data that should be changed.

For example, IEA publishes new wind turbine cost estimates and a modeller updates the data like
```python
inv_cost_wind = pd.DataFrame()  # get a dataframe of investment cost parameters for a certain technology
run.parameters.get("inv_cost").add_data(
  {‘technology’: ‘wind’, ‘region’: ‘EEU’, ‘value’: w, ‘unit’: ‘EUR/GW’}
)
```

glatterf42:
Okay, I've updated the behaviour to this:
```python
        table_3 = run.optimization.tables.create(
            name="Table 3",
            constrained_to_indexsets=[indexset.name, indexset_2.name],
            column_names=["Column 1", "Column 2"],
        )

        table_3.add(data={"Column 1": ["bar"], "Column 2": [2]})
        assert table_3.data == {"Column 1": ["bar"], "Column 2": [2]}

        # Test data is expanded when Column.name is already present
        table_3.add(
            data=pd.DataFrame({"Column 1": ["foo"], "Column 2": [3]}),
        )
        assert table_3.data == {"Column 1": ["bar", "foo"], "Column 2": [2, 3]}
```

If this is now how it should be, I'll add a DB migration and hopefully we can merge this PR soon :)

danielhuppmann:
Thanks @glatterf42. Just to be sure, continuing the example above would work like his:

```python
        # Test data is expanded when Column.name is already present
        table_3.add(
            data=pd.DataFrame({"Column 1": ["bar"], "Column 2": [1]}),
        )
        assert table_3.data == {"Column 1": ["bar", "foo"], "Column 2": [1, 3]}
```

danielhuppmann:
And I think the order of arguments in `table.create` should be "name -> column_names -> constrained_to_indexsets" for better readability.

ixmp4/data/api/optimization/column.py
```diff
@@ -0,0 +1,60 @@
...
+    abstract.ColumnRepository,
+):
+    model_class = Column
+    prefix = "optimization/columns/"
+
+    def __init__(self, client, *args, **kwargs) -> None:
+        super().__init__(client, *args, **kwargs)
+        self.docs = ColumnDocsRepository(self.client)
+
+    # TODO: This is not currently in use, only here because the abstract class has this
```
meksor, line 44: Just to clarify, not in use because the rest api side is not done or not in use because it is only used internally?
In the second case its probably better to omit ColumnRepository from the abstract interface entirely.

ixmp4/data/db/optimization/column/filter.py
```diff
@@ -0,0 +1,11 @@
+from ixmp4.data.db import filters as base
+from ixmp4.db import filters
+
+
+class OptimizationColumnFilter(
+    base.OptimizationColumnFilter, metaclass=filters.FilterMeta
+):
+    table: base.OptimizationColumnFilter | None
```
meksor, line 8: Is this a typo?

ixmp4/data/db/optimization/column/model.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    indexset: types.Mapped[IndexSet] = db.relationship(single_parent=True)
+    constrained_to_indexset: types.Mapped[int] = db.Column(
+        db.Integer, db.ForeignKey("optimization_indexset.id"), index=True
+    )
+
+    # Currently not in use:
+    unique: types.Boolean = db.Column(db.Boolean, default=True)
+
+    __table_args__ = (UniqueConstraint(name, table__id, constrained_to_indexset),)
```
meksor, line 35: Can there be two columns with the same names but constrained to different sets?
Not really, right? Probably need to adjust this constraint.

ixmp4/data/db/optimization/table/filter.py
```diff
@@ -0,0 +1,11 @@
+from ixmp4.data.db import filters as base
+from ixmp4.db import filters
+
+
+class OptimizationTableFilter(
+    base.OptimizationTableFilter, metaclass=filters.FilterMeta
+):
+    table: base.OptimizationTableFilter | None
```
meksor, line 8: The table filter has an item named table which is the same filter again?

tests/data/test_docs.py
```diff
@@ -328,3 +328,67 @@ def test_delete_scalardocs(self, test_mp, request):
 
         with pytest.raises(Docs.NotFound):
             test_mp.backend.optimization.scalars.docs.get(scalar.id)
+
+
+@database_platforms
+class TestDatabaseDocs:
```
meksor, line 334: Why is this only tested for databases?

glatterf42:
> Thanks @glatterf42. Just to be sure, continuing the example above would work like his:
> 
> ```python
>         # Test data is expanded when Column.name is already present
>         table_3.add(
>             data=pd.DataFrame({"Column 1": ["bar"], "Column 2": [1]}),
>         )
>         assert table_3.data == {"Column 1": ["bar", "foo"], "Column 2": [1, 3]}
> ```

Is this really what we want? To add some context, we of course have two indexsets with these elements defined before:
```python
indexset.add(elements=["foo", "bar"])
indexset_2.add([1, 2, 3])
```
However, this mean that every combination of ("bar", 1), ("bar", 2), and ("bar", 3) is a valid Key. So if you add ("bar", 1) to a table that already has ("bar", 2), would we really want that to be overwritten or shouldn't the table rather receive another Key, so that it would be
```python
assert table_3.data == {"Column 1": ["bar", "foo", "bar"], "Column 2": [2, 3, 1]}
```
?

tests/data/test_docs.py
glatterf42: Thanks, forgot to remove that after I added the other layers.

danielhuppmann:
My apologies @glatterf42, you are absolutely right - I was confused. Please disregard my previous statement.

ixmp4/data/db/optimization/column/model.py
glatterf42: Yes, @danielhuppmann mentioned above that only the names should be unique, so I'll drop `constrained_to_indexsets` from this `UniqueConstraint`.

ixmp4/data/api/optimization/column.py
glatterf42: Not in use because `Table`s only interact with `Column`s on the DB layer level, so I figured there's no need to add an `add_column()` function to `api.Table`, which would only call `api.Column.create()`, which in turn calls `db.Column.create()`. So I guess its kind of all of the above, but I'll just move the docstring to the function in the DB layer and remove the `ColumnRepository` from the abstract layer, as you suggested.

glatterf42:
> And I think the order of arguments in `table.create` should be "name -> column_names -> constrained_to_indexsets" for better readability.

This is currently not possible because `column_names` has a default value while `constrained_to_indexsets` has not (I think this would result in a linting error, though ruff doesn't immediately flag it for me). The reason for this is that we take the column names from the names of the indexsets in `constrained_to_indexsets` per default and only allow overwriting them with `column_names`. We could change our approach to this, but it makes sense for me to offer the option to provide names only once if they correspond to `Indexset.name`s.

ixmp4/data/db/optimization/table/filter.py
glatterf42: This was adapted from Scalar, which came from Indexset and at some point, it was probably necessary to have this in. I have to admit that I'm not entirely sure why the current way of writing these filters is better, but of course I'll update/clean up the optimization filters.

meksor: Not sure im completely follwing right now, in this specific case wouldn't you just remove the table: ... annotation?

glatterf42: Yes, I did, but there were other changes introduced by your recent #56. See e.g. https://github.com/iiasa/ixmp4/blob/bcbdff3522f7d61be4d0d21060336238199bd0ed/ixmp4/data/db/optimization/indexset/filter.py#L8-L19 and compare to [how it was before](https://github.com/iiasa/ixmp4/blob/7798462920e7f06167163d3380f092c310581ae2/ixmp4/data/db/optimization/indexset/filter.py). I've just made the same changes now for all Optimization*Filters assuming that they represent a performance improvement.

meksor (approved):
Hiya, I looked at the current state of this PR code-wise and it looks good.

ixmp4/core/optimization/table.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class TableRepository(BaseFacade):
+    _run: Run
+
+    def __init__(self, _run: Run, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._run = _run
+
+    # TODO Here, constrained_to_indexsets is a list[str], while above it's list[int]
```
danielhuppmann, line 90: I'd have a strong preference to have indexsets identified by name (str)

ixmp4/core/optimization/table.py
```diff
@@ -0,0 +1,121 @@
...
+    # We should make this consistent
+    def create(
+        self,
+        name: str,
+        constrained_to_indexsets: list[str],
+        column_names: list[str] | None = None,
```
danielhuppmann, line 96: Shouldn't this be flipped around? column-names are required, constrained_to_indexsets is optional? And also changing the order would be more intuitive.

```suggestion
        column_names: list[str],
        constrained_to_indexsets: list[str] | None = None,
```

ixmp4/data/api/optimization/column.py
```diff
@@ -0,0 +1,24 @@
...
+
+    id: int
+    name: str
+    dtype: str
+    table__id: int
+    indexset: IndexSet
+    constrained_to_indexset: int
+    unique: bool
+
+    created_at: datetime | None
```
danielhuppmann, line 23: If the columns are created at the same time as the Table, I don't think that we need to have the created-info.

tests/core/test_table.py
```diff
@@ -0,0 +1,275 @@
...
+
+        table_2 = run.optimization.tables.create(
+            name="Table 2",
+            constrained_to_indexsets=[indexset.name, indexset_2.name],
+        )
+
+        with pytest.raises(ValueError, match="missing values"):
+            _ = table_2.add(
+                pd.DataFrame({"Indexset": [None], "Indexset 2": [2]}),
+                # empty string is allowed for now, but None or NaN raise
```
danielhuppmann, line 141: can you please add a test for the empty string, just to be sure that this actually works?

danielhuppmann (commented):
Thank you @glatterf42, this looks very nice, almost ready to be merged. A few observations inline, and one more general observation (which I made a while ago already): I believe that the order of arguments `table-name -> table-columns -> table-columns-constraints-to-indexsets` would be more intuitive...

ixmp4/core/optimization/table.py
glatterf42: This was resolved already, just forgot to remove the `TODO` :)

ixmp4/core/optimization/table.py
glatterf42: To quote my reply from when you first brought this up:

> This is currently not possible because `column_names` has a default value while `constrained_to_indexsets` has not (I think this would result in a linting error, though ruff doesn't immediately flag it for me). The reason for this is that we take the column names from the names of the indexsets in `constrained_to_indexsets` per default and only allow overwriting them with `column_names`. We could change our approach to this, but it makes sense for me to offer the option to provide names only once if they correspond to `Indexset.name`s.

Please keep in mind that we are currently requiring every `Column` of a `Table` to be linked to an `IndexSet` for validation. Your suggestion here changes the name of the arguments, so we would require an argument called `column_names` and optionally take `constrained_to_indexsets` as well. We would then presumably try to get an `IndexSet` for each name in `column_names` and link it to the `Column`. If we don't find any `IndexSet` with that name, we would suddenly require `constrained_to_indexsets` to be there and provide existing names for every `Column` (unless we change that to be a mapping, which might be a good idea). 
In the current approach, we require `constrained_to_indexset` to always be there with existing `IndexSet.name`s, but this is a transparent requirement: it's just always required. Per default, we use the `IndexSet.name`s for the `Column` names, too, but you can still overwrite them if you prefer. 

I don't know, the current approach is more intuitive to me.

danielhuppmann: Thanks for explaining it again, all good to proceed as is.

danielhuppmann (approved):
Looks good to me, many thanks!

PR #42, opened 2024-01-31
Upload Chunking

meksor:
Implements the chunking of bulk uploads and deletes in the api layer. Chunk size can be configured with `settings.default_upload_chunk_size` or `IXMP4_DEFAULT_UPLOAD_CHUNK_SIZE`. 

**NOTE**: If an upload is interrupted between chunks the operation is **not rolled back**, so use it with care until the versioning system is in place. 

I've also added some middleware to log the request size and process time. Seems like most request bodies are now max. around 1 Megabyte so i will probably set the webserver config to 2M max body size after this change is deployed (current is 50M). 
IXMP4 will also now throw a verbose error if the webserver responds with a 413 (Payload too large).

### DIFF
--- a/ixmp4/conf/settings.py
+++ b/ixmp4/conf/settings.py
@@ -34,7 +34,7 @@ class Settings(BaseSettings):
     managed: bool = True
     max_page_size: int = 10_000
     default_page_size: int = 5_000
-
+    default_upload_chunk_size: int = 10_000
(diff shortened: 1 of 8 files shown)


meksor:
PS: I bumped fastapi to `0.108.0` to make the size logging happen.

glatterf42 (approved):
Looks good to me :)

danielhuppmann (approved):
Nice!

PR #43, opened 2024-02-06
Fix dynamic versioning

meksor:
Was pretty simple: the plugin runs some simple string substitution over all files, making it work by just using these two lines in the top level `__init__.py`:

```py
__version__ = "0.0.0"
__version_tuple__ = (0, 0, 0)
```


danielhuppmann (approved):
Nice!

PR #46, opened 2024-02-07
Fix filters only being callable once

meksor:
Weirdly, when calling model_dump twice on the same pydantic model instance, the second time all fields defined with "Iterable" are empty...
Added a fix to circumvent this.


glatterf42 (approved):
Tests are passing, so looks good to me. Do we have a test that confirms filters are now callable more than once?

meksor:
You're right, i added some tests for pagination specifically and found another bug, thanks!

PR #47, opened 2024-02-09
Docker Builds on Publish

meksor:
Adds a github action which is supposed to build containers on tags-pushed or publish.


meksor:
OK, phew that took a while to get right. 
To ensure compatibility with the broken docker version on some of the IIASA servers, we need to pin the buildkit moby image to version v0.10.5 and disable pip progress bars. 
Anyway, full releases are tagged with the version string and "latest" while pushed git tags are taggs with the version string and "dev" (similar to the pypi and test-pypi uploads).

glatterf42 (changes requested):
One minor change, but looks good otherwise.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10
+FROM python:3.11
```
glatterf42, line 1:
```suggestion
FROM python:3.12
```
Just wondering: if we merge #41 first, this should probably become 3.12 too, right?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,48 @@
...
+          version: latest
+          driver-opts: "image=moby/buildkit:v0.10.5"
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          registry: ene-docker.iiasa.ac.at
+          username: ixmp-server-workflow-build-user
+          password: ${{ secrets.NEXUS_PASSWORD }}
+      - name: Check out repository
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
meksor: thanks

Dockerfile
meksor: I guess we can try!

glatterf42 (approved):
Looks good to me now, thanks :)

PR #51, opened 2024-02-28
Fix wrong column used in datapoint filter

meksor:
Also: disable debug logging for watchfiles (as it spams the debug log).


danielhuppmann (approved):
Looks good to me, thanks!

PR #53, opened 2024-02-29
Add extra check if both values are NA in bulk_upsert

### LINKED ISSUE #50: Version 0.7.0 breaks pyam test
As I was working on pyam yesterday (IAMconsortium/pyam#818) I noticed that `ixmp4 0.7.0` broke the test `test_ixmp4_integration[test_df0]` from `tests/test_ixmp4.py` (https://github.com/IAMconsortium/pyam/actions/runs/8067457546/job/22037916444) with the following error:
 
```console
...
pyarrow/error.pxi:154: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowNotImplementedError: Function 'not_equal' has no kernel matching input types (string, double)

pyarrow/error.pxi:91: ArrowNotImplementedError
```

looks like it's got something to do with `pyarrow`.
Reverting to ixmp4 version 0.6.0 fixed the test.
Just talked to @meksor and he said that you, @glatterf42, would be the best person to take a look.

FYI @danielhuppmann.

meksor:
Fixes #50 's pyarrrow compatibility problem.


ixmp4/data/db/base.py
```diff
@@ -391,7 +391,10 @@ def bulk_upsert_chunk(self, df: pd.DataFrame) -> None:
             for col in self.model_class.updateable_columns:
                 updated_col = col + self.merge_suffix
                 if updated_col in df.columns:
-                    cond.append(df[col] != df[updated_col])
+                    df[updated_col] = df[updated_col].astype(df[col].dtype)
```
danielhuppmann, line 394: Maybe add a line explaining why this is necessary? And a test (in a GitHub Action that explicitly installs pyarrow) to guard against regression?

glatterf42:
Do we intend to release 0.7.1 immediately after this fix so that https://github.com/IAMconsortium/pyam/pull/820 can move forward?

glatterf42 (approved):
Looks good to me, provided the tests pass. But I'm not sure I like  test names like "test / test (3.11, false) (pull_request) ". Doesn't seem as descriptive as it could be, but I don't have a tangible better solution of the top of my head.

PR #54, opened 2024-03-05
add iamc filter for tabulate/ list runs

pmussak:
Added `iamc` filter similar to `mp.backend.models.tabulate` and `mp.backend.scenarios.tabulate` to `mp.runs.tabulate` (all applies to `.list` methods as well).

Why was this added?
To be able to filter runs by whether they have `IAMC` data associated with certain properties, e.g. get all runs that have `IAMC` data for the `Cumulative Emissions|CO2` variable or the `World` region.

usage:
```
mp.runs.tabulate(
            iamc={
                "variable": {"name__like": "Cumulative Emissions|CO2*"},
                "unit": {"name": "World"},
            }
        )
```

### DIFF
--- a/tests/core/test_run.py
+++ b/tests/core/test_run.py
@@ -96,7 +96,7 @@ class TestCoreRun:
 
     def test_filter_run(self, test_mp, request):
         test_mp = request.getfixturevalue(test_mp)
-        run1, _ = create_filter_test_data(test_mp)
+        run1, run2 = create_filter_test_data(test_mp)
 
         res = test_mp.runs.tabulate(
             default_only=False,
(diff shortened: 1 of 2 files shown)


pmussak:
@danielhuppmann please take a look on whether this is sufficient for the needs for filtering runs on IAMC data that you perceived (for ixmp4-ts it is sufficient).

glatterf42 (approved):
Thanks, looks good to me :)

@meksor might want to take a look as well.

meksor (approved):
looks good!

tests/core/test_run.py
```diff
@@ -93,3 +93,47 @@ def test_run_versions(self, test_mp, request):
...
+        assert sorted(res["model"].tolist()) == ["Model 2"]
+        assert sorted(res["scenario"].tolist()) == ["Scenario 2"]
+
+        res = test_mp.runs.tabulate(
+            default_only=False,
+            scenario={"name": "Scenario 2"},
+        )
+
+        assert sorted(res["model"].tolist()) == ["Model 2"]
+        assert sorted(res["scenario"].tolist()) == ["Scenario 2"]
```
danielhuppmann, line 139: Suggest to extend the test in the following way:
- Remove all datapoints for the run Model 1, Scenario 1 related to Region 3
- Repeat the assertion form above
  ```python
           res = test_mp.runs.tabulate(
             default_only=False,
             iamc={
                 "region": {"name": "Region 3"},
             },
         )
   ```
- This should now only return Model 2, Scenario 2

danielhuppmann (commented):
Thanks for adding this additional filter (I had thought that this was already implemented)...

One clarifying question: this filter will only return items where a datapoint exists? (not just a measurand linked to the run)

See a possible extension inline.

tests/core/test_run.py
pmussak: @danielhuppmann thanks for raising this. I added a test as you proposed and only runs with existing datapoints are returned :+1:

danielhuppmann (approved):
Thanks, looks good to me!

PR #59, opened 2024-03-08
Fetch tags in CI for correct version

glatterf42:
I noticed before that our CI is not installing the correct version of our own library (it keeps being registered as v0.0.0). While reading up on [poetry-dynamic-versioning limitations](https://pypi.org/project/poetry-dynamic-versioning/), I found a [note about Dunamai's limitations](https://github.com/mtkennerly/dunamai#other-notes), which powers poetry-dynamic-versioning. In order to register the version correctly in GitHub Actions, they recommend running the checkout action with `fetch-depth: 0`, which fetches all history for all branches and tags. However, [v4 of the checkout action](https://github.com/actions/checkout) also offers `fetch-tags`, which is `false` by default. 
In theory, this sounds exactly like the thing we need, so this PR sets it to `true` in the relevant workflows.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -93,6 +93,8 @@ jobs:
       #       check-out repo and set-up python
       #----------------------------------------------
       - uses: actions/checkout@v4
+        with:
+          fetch-tags: true
       - uses: actions/setup-python@v5
       #----------------------------------------------
       #  -----  install & configure poetry  -----
@@ -103,6 +105,11 @@ jobs:
           virtualenvs-create: true
           virtualenvs-in-project: true
           installer-parallel: true
+      #----------------------------------
+      #  install dynamic version plugin  
+      #----------------------------------
+      - name: Install poetry-dynamic-versioning
+        run: poetry self add "poetry-dynamic-versioning[plugin]"
       #----------------------------------------------
       #       load cached venv if cache exists
       #----------------------------------------------
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v4
         with:
-          fetch-tags: true
+          fetch-depths: 0
       - name: Set up python
         id: setup-python
         uses: actions/setup-python@v5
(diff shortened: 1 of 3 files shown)


glatterf42:
The simple `fetch-tags: true` did not fetch any tags, it seems. However, even though `fetch-depth: 0` did fetch all tags (check the build-docs run from the second commit), the installed version of ixmp4 remains 0.0.0.

glatterf42:
The solution was already partly contained in the publish workflow: we need to install the dynamic-versioning plugin in the CI runs as well, of course. 
However, the old publish workflow still produced a warning in [the build package step](https://github.com/iiasa/ixmp4/actions/runs/8171554784/job/22340027070): `Warning: This is a shallow repository, so Dunamai may not produce the correct version.`
This PR should remedy this by cloning the whole repo history.

One final note: we are currently only using lightweight tags. [Git also offers annotated tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging), which save the tag author and tag date. Should we ever apply multiple lightweight tags to the current commit, poetry-dynamic-versioning might struggle to identify the most recent tag. The solution for this is to use annotated tags. We can create them easily via `git tag -a v0.7.3` (for example). This will ask for a tag creation message, which will be displayed when investigating the tag e.g. via `git show v0.7.3`. It might also be possible to use `git tag -s`, which gpg-signs a tag, to create an annotated tag. I'm not sure it this also requires a tag message.

.github/workflows/pytest.yaml
```diff
@@ -101,6 +110,11 @@ jobs:
           virtualenvs-create: true
           virtualenvs-in-project: true
           installer-parallel: true
+      #----------------------------------
```
danielhuppmann, line 113: This heading is less wide than the others

.github/workflows/pytest.yaml
```diff
@@ -91,6 +98,8 @@ jobs:
       #       check-out repo and set-up python
       #----------------------------------------------
       - uses: actions/checkout@v4
+        with:
+          fetch-depths: 0
       - uses: actions/setup-python@v5
       #----------------------------------------------
       #  -----  install & configure poetry  -----
```
danielhuppmann, line 105: The dashes are different only for the "install & configure poetry" section headers.

danielhuppmann (approved):
Thanks, looks good - but please go through the workflow files once more and make the section-headings consistent (centered and see other comments inline).

PR #61, opened 2024-03-13
Add needed [dataframe] extra for dask

glatterf42:
Downstream in message_ix and ixmp (among others), CI tests using the latest dask version (2024.3.0) are failing because ["dask-expr"](https://github.com/dask/dask-expr) is missing. This is the new backend for `dask.DataFrame`s and is required because of ixmp4, which doesn't currently specify any extra for dask, thus missing "dask-expr". This PR introduces the "dataframe" extra for dask, which installs "dask-expr". 

I also ran `poetry lock` in the process, so some minor version updates may be included in `poetry.lock`.
This additionally threw an error message which was related to pandas' coming [Copy-on-Write default behaviour](https://pandas.pydata.org/pandas-docs/stable/user_guide/copy_on_write.html#copy-on-write), so I quickly fixed that. However, I should open another issue to keep track of this change in pandas and see how it affects us. 

@danielhuppmann @meksor (@pmussak FYI) Since downstream projects depend on this fix, could we please do a release 0.7.3 today already?


danielhuppmann (approved):
Thanks @glatterf42

PR #62, opened 2024-03-13
update DEVELOPING.md

meksor:
Added some info about the benchmarks. @glatterf42


glatterf42 (approved):
Thanks, lgtm :)
Though I will only see at a later time if the instructions are clear enough for me to actually compare performance of ixmp and ixmp4.

PR #64, opened 2024-03-13
Implement simple filters

danielhuppmann:
This PR implements a "short-hand notation" for filters in the facade-layer following the conventions of the pyam package, which allows to do `df.filter(variable="Primary *")` for like-comparing (using * as a wildcard).

The implementation replaces string or lists to the more elaborate filter-dictionary-style used in the ixmp4 data layer and the Rest API.


meksor (approved):
lgtm

PR #65, opened 2024-03-14
Fix RTD builds with poetry

glatterf42:
Thanks @danielhuppmann for noticing our builds were failing and finding the solution over at https://github.com/IAMconsortium/pyam/pull/836 :)


danielhuppmann (approved):
Thanks @glatterf42!

glatterf42:
For completeness, successful build is here: https://docs.ece.iiasa.ac.at/projects/ixmp4/en/fix-build-docs/

Issue #55, opened 2024-03-05
HTTP 500 on `/v1/genie/meta/`


pmussak:
I am using the `meta` REST endpoint in`ixmp4-ts`. While developing I found that,
1. calls for `https://ixmp.ece.iiasa.ac.at/v1/genie/meta/?join_run_index=true&table=true`
result in HTTP 500 **
context: current prod deployment of ixmp4 (0.7.1)

2. calls for `https://ixmp.ece.iiasa.ac.at/v1/genie/meta/?table=true`
work as expected

3. calls for `https://ixmp.ece.iiasa.ac.at/v1/genie/meta/?join_run_index=true`
result in HTTP 400 ***

I expect that all three requests result in HTTP 2** responses with data in the result.

** ixmp4 error log for first request
```
sqlalchemy.exc.ProgrammingError: (psycopg.errors.DuplicateAlias) table name "model" specified more than once
[SQL: SELECT DISTINCT model.name AS model, scenario.name AS scenario, run.version, runmetaentry.run__id, runmetaentry.key, runmetaentry.type, runmetaentry.value_int, runmetaentry.value_str, runmetaentry.value_float, runmetaentry.value_bool, runmetaentry.id 
FROM runmetaentry JOIN run ON run.id = runmetaentry.run__id JOIN model ON model.id = run.model__id JOIN model ON model.id = run.model__id JOIN scenario ON scenario.id = run.scenario__id 
WHERE model.name LIKE %(name_1)s::VARCHAR AND run.is_default AND run.is_default = true AND model.name LIKE %(name_2)s::VARCHAR AND run.is_default AND run.is_default = true ORDER BY runmetaentry.id ASC 
 LIMIT %(param_1)s::INTEGER OFFSET %(param_2)s::INTEGER]
[parameters: {'name_1': '%', 'name_2': '%', 'param_1': 5000, 'param_2': 0}]
(Background on this error at: https://sqlalche.me/e/20/f405)
```

*** error response for third request
```
{
    "message": "The provided filter arguments are malformed.",
    "kwargs": {
        "model": "run",
        "errors": [
            {
                "type": "extra_forbidden",
                "loc": [
                    "join_run_index"
                ],
                "msg": "Extra inputs are not permitted",
                "input": true,
                "url": "https://errors.pydantic.dev/2.6/v/extra_forbidden"
            }
        ]
    },
    "error_name": "bad_filter_arguments"
}
```

pmussak:
[...]

PR #66, opened 2024-03-16
Add filtering by model and scenario name in MetaRepository

danielhuppmann:
For the integration with **pyam**, I want a feature to filter the MetaRepository by model and scenario name, like
```python
platform.meta.tabulate(run={'default_only': True}, scenario={name="scen_a"})
```

The PR below implements these filters and they work as expected, but I get the following warning
```
/Users/dh/.venv/iamc/lib/python3.12/site-packages/pandas/io/sql.py:1600: SAWarning:
SELECT statement has a cartesian product between FROM element(s) "runmetaentry"
and FROM element "scenario".  Apply join condition(s) between each element to resolve.
```

Any pointers to how/where to resolve that, @meksor @pmussak?

### DIFF
--- a/ixmp4/core/meta.py
+++ b/ixmp4/core/meta.py
@@ -5,8 +5,7 @@ from .base import BaseFacade
 
 class MetaRepository(BaseFacade):
     def tabulate(self, **kwargs) -> pd.DataFrame:
-        # TODO: accept list of `Run` instances as arg
-        # TODO: expand run-id to model-scenario-version-id columns
+        # TODO: accept list of `Run` instances as argument
         return self.backend.meta.tabulate(join_run_index=True, **kwargs).drop(
             columns=["id", "type"]
         )
--- a/ixmp4/data/db/meta/filter.py
+++ b/ixmp4/data/db/meta/filter.py
@@ -1,24 +1,10 @@
-from typing import ClassVar
-
 from ixmp4.data.db import filters as base
-from ixmp4.data.db.model import Model
 from ixmp4.data.db.run import Run
-from ixmp4.data.db.scenario import Scenario
+from ixmp4.data.db.model.filter import ModelFilter
+from ixmp4.data.db.scenario.filter import ScenarioFilter
 from ixmp4.db import filters, utils
 
 from .model import RunMetaEntry
-
-
-class ModelFilter(base.RunFilter, metaclass=filters.FilterMeta):
-    name: filters.String
-
-    sqla_model: ClassVar[type] = Model
-
-
-class ScenarioFilter(base.RunFilter, metaclass=filters.FilterMeta):
-    name: filters.String
-
-    sqla_model: ClassVar[type] = Scenario
 
 
 class RunFilter(base.RunFilter, metaclass=filters.FilterMeta):
--- a/ixmp4/data/db/meta/filter.py
+++ b/ixmp4/data/db/meta/filter.py
@@ -1,8 +1,24 @@
+from typing import ClassVar
+
 from ixmp4.data.db import filters as base
+from ixmp4.data.db.model import Model
 from ixmp4.data.db.run import Run
+from ixmp4.data.db.scenario import Scenario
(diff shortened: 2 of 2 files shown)


meksor:
Wait a minute, shouldn't filtering via model/scenario already be possible with the run filter? 
Like: `run={"model": {"name":...}}`

meksor:
Also both your Model and Run filters inherit from "RunFilter" which is probably the reason for the warning. If the join paths are incorrect the database will sometimes try to calculate the cartesian product of two tables. 
I don't know why it does this, I haven't ever encountered a use case for this but maybe it's a historical artifact of the computer science background of databases. 
(Outer joins are expressed as the cross product of two tables in some databases too)

danielhuppmann:
Thanks @meksor for the pointer.

I believe that 
```python
platform.meta.tabulate(run={model="Model 1"})
```
is deactivated by https://github.com/iiasa/ixmp4/blob/3f45bba201de4a448f5e7be930b9e6593fb59d49/ixmp4/data/db/meta/filter.py#L16

And after some back and forth, I think that the API should indeed be 
```python
platform.meta.tabulate(run={default_only=True}, model="Model 1"})
```
because the model-name is an attribute of the Model table, not the Run table.

This is similar to the (already-working) filter-signature of the iamc datapoints:
```python
platform.iamc.tabulate(run={'default_only': True}, model="Model 1")
```

danielhuppmann:
In the commit above, I switched to import
```python
from ixmp4.data.db.model.filter import ModelFilter
```
which yields the same warning (and also works as expected).

I also tried 
```python
from ixmp4.data.db.filters.model import ModelFilter
```
which raised an error.

I have to admit that the difference between `model.filter` and `filters.model` is still not clear to me...

I guess I would need to do something like
```
WHERE run.d in (SELECT run_id from SCENARIO WHERE name == ...)
```
instead of the multi-join...?

meksor:
> Thanks @meksor for the pointer.
> 
> I believe that 
> ```python
> platform.meta.tabulate(run={model="Model 1"})
> ```
> is deactivated by https://github.com/iiasa/ixmp4/blob/3f45bba201de4a448f5e7be930b9e6593fb59d49/ixmp4/data/db/meta/filter.py#L16
> 

? I don't think that's right

meksor:
The run filter was one of the first filters in ixmp, you can always filter by a model and its name through the runs filter except on the /iamc/datapoints filter. Because its supposed to act like that big table i guess...
I would think about moving this stuff around now, seems like we are introducing even more custom special-case inconsistencies...

meksor:
It bad enough IMO that the defaults for the filter are manually set every time it is used, that seems horrible to be, but whatever ^^

danielhuppmann:
Not quite sure what you mean by "not right", but from my perspective

```
test.mp.runs.tabulate()
```
should accept filters `default_only`, `run_id`, `version`, `model` (name and id) and `scenario` (name and id),

but a `RunFilter` used as filter from iamc-datapoints or meta should only accept `default_only`, `run_id` and `version` because these are the attributes in the Run table.

So I guess we have to split the RunFilter used in `Runs` from the other uses? That would also allow not having to set the defaults every time...

meksor:
So i just double checked, and indeed, we can already filter by models with, for example 
```json
{
  "run": {
    "model": {"name__like": "Model *"},
    "default_only": true,
    "is_default": true
  }
}
```
And presumably that enables: `tabulate(run={"model": {"name": "..."}})

danielhuppmann:
Your presumption is correct for `platform.runs.tabulate()` but not for `platforms.meta.tabulate()`...

meksor:
Ah, but i think that is because of this issue: https://github.com/iiasa/ixmp4/issues/55
Not because it is not implemented like that.

meksor:
Well then it looks like you found a workaround for the issue...
I still think it would work if that issue is fixed, im just getting a 500 when i try on the current ixmp instances.

danielhuppmann:
Closing in favor of #69

PR #69, opened 2024-03-18
Fix Run Meta Query Building

### LINKED ISSUE #55: HTTP 500 on `/v1/genie/meta/`

meksor:
Fixes query building for run meta entry queries. Also adds CI tests for postgres 15 and 16. 
Throws a 400 when wrong parameters are used on the meta endpoint.
I also added a compatibility matrix to README.md.
Fixes #55.
(Maybe) implements [PR#66](https://github.com/iiasa/ixmp4/pull/66)?

### DIFF
--- a/ixmp4/data/db/run/repository.py
+++ b/ixmp4/data/db/run/repository.py
@@ -1,7 +1,5 @@
 import pandas as pd
-from sqlalchemy import select
 from sqlalchemy.exc import NoResultFound
-from sqlalchemy.orm import Bundle
 
 from ixmp4 import db
 from ixmp4.core.exceptions import Forbidden, IxmpError, NoDefaultRunVersion
--- a/ixmp4/data/db/run/repository.py
+++ b/ixmp4/data/db/run/repository.py
@@ -176,17 +176,3 @@ class RunRepository(
 
         run.is_default = False
         self.session.commit()
-
-
-def select_joined_run_index(repository, **kwargs) -> db.sql.Select:
-    _exc = select(
-        Bundle("Model", Model.name.label("model")),
-        Bundle("Scenario", Scenario.name.label("scenario")),
-        Bundle("Run", Run.version),
-        repository.bundle,
-    )
-    return (
-        repository.select(_exc=_exc, **kwargs)
-        .join(Model, onclause=Model.id == Run.model__id)
-        .join(Scenario, onclause=Scenario.id == Run.scenario__id)
-    )
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -13,13 +13,18 @@ jobs:
           - "3.12" # Latest version supported by ixmp4
         with-pyarrow:
           - false
+        postgres-version:
+          - 15
         include:
           - python-version: "3.12"
             with-pyarrow: true
+          - python-version: "3.12"
+            with-pyarrow: false
+            postgres-version: 16
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres
+        image: postgres:${{ matrix.postgres-version }}
         env:
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: test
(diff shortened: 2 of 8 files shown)


ixmp4/data/db/run/repository.py
```diff
@@ -176,17 +174,3 @@ def unset_as_default_version(self, id: int) -> None:
 
         run.is_default = False
         self.session.commit()
-
-
-def select_joined_run_index(repository, **kwargs) -> db.sql.Select:
```
danielhuppmann, line 181: For what it's worth, I added this function here so that I could use it later to streamline the following almost-duplicate code parts:

https://github.com/iiasa/ixmp4/blob/3f45bba201de4a448f5e7be930b9e6593fb59d49/ixmp4/core/run.py#L94

https://github.com/iiasa/ixmp4/blob/3f45bba201de4a448f5e7be930b9e6593fb59d49/ixmp4/data/db/iamc/datapoint/repository.py#L111

danielhuppmann (commented):
Thanks @meksor for digging through the code to fix this issue!

ixmp4/data/db/run/repository.py
meksor: Yeah i think id take a look at this again in some sort of selection overhaul PR. The problem is that depending on what table is selected the joins have to be different (and there was a bug in the code applying filters twice) so i would try to find some new abstraction for this and the filters to make it less error prone in the future.

meksor: I also had to rename the columns selected ("model"->"model_name") because postgres 15 does not like columns in the result set with the same name as a table from the query

danielhuppmann (approved):
Looks good to be merged and have the filters-signature discussion later…

PR #70, opened 2024-03-18
Fix API deletes

### LINKED ISSUE #57: Deleting the docs via Rest fails on a production instance
Per the tests, deleting the docstring of a region should work using 
```python
region.docs = None
```
or via the explicit `delete_docs` method, see [here](https://github.com/iiasa/ixmp4/blob/6e04ed3896fbf0e86f9346dc7a5988c320f5d1b4/tests/core/test_region.py#L128).

However, this raises the following error on the ecemf-internal instance:

image

meksor:
Currently on the production setup deletes fail because no trailing slash is added in the api layer and there is some mishap in how that request is redirected...
I've fixed the api layer and set `redirect_slashes=False` for fastapi. All requests to endpoints without trailing slashed will thus result in a `404`. 
The docs are now also served with trailing slashes so i can use an nginx rewrite instead of the fastapi redirect.
Should fix #57

### DIFF
--- a/ixmp4/data/api/base.py
+++ b/ixmp4/data/api/base.py
@@ -232,7 +232,7 @@ class BaseRepository(Generic[ModelType]):
         return self._request("POST", *args, **kwargs)  # type: ignore
 
     def _delete(self, id: int):
-        self._request("DELETE", self.prefix + str(id))
+        self._request("DELETE", f"{self.prefix}{str(id)}/")
 
 
 class Retriever(BaseRepository[ModelType]):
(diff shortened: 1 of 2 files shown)


glatterf42 (approved):
Looks good to me :)

Why do we need to set `docs_url` and `redoc_url` specifically?

meksor:
> Why do we need to set `docs_url` and `redoc_url` specifically?

Because otherwise they are served at a location without a trailing slash.

PR #74, opened 2024-03-20
Fix slow run filter

meksor:
The performance improvement from [PR#56](https://github.com/iiasa/ixmp4/pull/56) didnt make it into the run filter. 
This PR intends to fix that and fix some behaviour of the iamc filter as well..

### DIFF
--- a/ixmp4/data/db/run/filter.py
+++ b/ixmp4/data/db/run/filter.py
@@ -1,5 +1,5 @@
+from ixmp4 import db
 from ixmp4.data.db import filters as base
-from ixmp4.data.db.iamc.datapoint import get_datapoint_model
 from ixmp4.data.db.iamc.timeseries import TimeSeries
 from ixmp4.data.db.run.model import Run
 from ixmp4.db import filters, utils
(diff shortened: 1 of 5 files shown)


ixmp4/data/db/run/filter.py
```diff
@@ -13,15 +13,27 @@ class IamcRunFilter(filters.BaseFilter, metaclass=filters.FilterMeta):
...
 class RunFilter(base.RunFilter, metaclass=filters.FilterMeta):
-    iamc: IamcRunFilter | filters.Boolean
+    iamc: IamcRunFilter | filters.Boolean | None = None
+
+    def join_datapoints(self, exc: db.sql.Select, session=None):
+        if not utils.is_joined(exc, TimeSeries):
+            exc = exc.join(TimeSeries, onclause=TimeSeries.run__id == Run.id)
+        return exc
+
+    def filter_iamc(self, exc, c, v, session=None):
```
glatterf42, line 27: I don't see `c` being used anywhere in the function, is that correct? 
Why don't we typehint `exc, c, v` to make more explicit what we expect here?

glatterf42 (commented):
Looks good to me in principle, just one comment.

ixmp4/data/db/run/filter.py
meksor: Happy to review or write a cleanup PR.
These functions are never type hinted, but i would adjust them /all/ while we are at it, not just this one.

danielhuppmann (approved):
Thanks @meksor!

PR #75, opened 2024-03-21
Clarify CI test names

glatterf42:
#69 added another dimension to the pytest CI runs, `postgres-version`. Our CI names now look like this: `test / test (3.10, false, 15) (pull_request)`. This PR adapts the names of the runs to add clarity about what the numbers and bool in the test name mean.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -22,7 +22,7 @@ jobs:
           - python-version: "3.12"
             with-pyarrow: false
             postgres-version: "16"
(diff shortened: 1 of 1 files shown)


meksor (approved):
Lgtm

glatterf42:
_Does_ it look good? I'm still wondering if we shouldn't add more whitespace to avoid the cluttered appearance.

meksor:
I mean, come on, it's a string describing a test run, not written aesthetic word-jazz. But if you want to improve further, at your leisure ^^

PR #76, opened 2024-03-26
Asyncronous Client

meksor:
Makes some of the client code asynchronous. This enables the concurrent sending of requests for enumerations. 
Asynchronicity is /not/ exposed to the user via async functions.
Adds some settings variables to control client behaviour, like retries, how long to wait before retrying and when the new `ApiEncumbered` exception is thrown.

```python
client_max_concurrent_requests: int = Field(2, lt=4)
client_max_request_retries: int = Field(3)
client_backoff_factor: int = Field(5)
client_timeout: int = Field(30)
```

Uses `asyncio.Semaphore` to control the maximum concurrency.


pyproject.toml
```diff
@@ -117,6 +116,9 @@ ignore = ["B008"]
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]
+# C901: warning for "too much branching logic"
+# in a function ...
+"ixmp4/data/api/base.py" = ["C901"]
```
glatterf42, line 121: Would we really rather exclude this file here than make the function less complex? Usually, moving one or two if-clauses to a helper function is enough to remedy this warning.

glatterf42 (commented):
Could you still add a test for the new functions, maybe especially if the new error is raised correctly? 

Other than that, looks good to me :)

pyproject.toml
meksor: You're right, I removed that explicit exception and split up the response handling from that "complex" function. 
Kind of makes it harder to read imo but there you go!

meksor:
I can't really add tests for that, no. There is in general 2 ways the new exception is triggered and they are both dependant on the webserver configuration and how beefy the setup is. 
For example, ixmp4 would never send the response code `429` because it is not able to tell if the entire system is overloaded. Only the webserver/proxy (in our case nginx) in front of ixmp4 or the process manager/ orchestration thingy (in our case docker/kubernetes) could know this and react to it by returning `429`. Also tracking IP's and rate limiting by IP is something best done in the webserver. 
I can imagine writing system tests that include ixmp4 and check if the exception is thrown under certain circumstances, but there is some way to go until then.

glatterf42 (approved):
Perfect, lgtm then :)

PR #77, opened 2024-03-26
Target PG16

meksor:
Since ICT has changed plans and installed pg version 16 for us, lets switch back to targeting that version mainly.


danielhuppmann:
Is that the reason why the nomenclature-nightly-tests failed?
See https://github.com/IAMconsortium/nomenclature/actions/runs/8610693572/job/23596636559

danielhuppmann (approved):
Looks good to me, thanks!

meksor:
Looks like thats the reason, yes!

PR #78, opened 2024-04-11
Remove Dask and Internal Validation

meksor:
Provides some performance improvements by removing dask (which was causing some overhead in the benchmarks, not sure when that changed) and some internal validation. 
Functions with the @check_types validator can now take an optional argument `skip_validation` if they are used internally to avoid unnecessary pandera overhead.
Both of these adjustments made things almost 70% faster on some datasets!


glatterf42 (commented):
This sounds super useful, though it will create a conflict with #68. And the conflict needs to be resolved, of course. Lgtm afterwards :)

glatterf42:
Almost, now it says there's a conflict in poetry.lock.

meksor:
Yeah because of my other PR :)

glatterf42 (approved):
Lgtm now :)

PR #79, opened 2024-04-15
Include optimization parameter basis

glatterf42:
As promised in a recent team meeting, the next steps for the message_ix data model work will use a more trunk-based development approach. This is where I got to today with introducing `optimization.Parameter`s. The tests for creating and getting them pass for the DB layer, others are still disabled. Ruff might complain that some temporarily-commented lines in the test file are now too long. 
If the tests run at all, maybe they only trigger on PRs to main. I could enable them for PRs to include/optimization-parameter as well or we merge this one and then set up a soon-to-be-final PR to main to collect further changes. 

I'm not yet sure this is the best way to handle the `values` and `units`: as separate lists. I will test if this behaves as desired/expected and will rework if it doesn't.

### DIFF
--- a/ixmp4/__init__.py
+++ b/ixmp4/__init__.py
@@ -1,16 +1,16 @@
-# flake8: noqa
 import importlib.metadata
 
+from ixmp4.core import IndexSet as IndexSet
 from ixmp4.core import Model as Model
+from ixmp4.core import Parameter as Parameter
 from ixmp4.core import Platform as Platform
 from ixmp4.core import Region as Region
 from ixmp4.core import Run as Run
+from ixmp4.core import Scalar as Scalar
 from ixmp4.core import Scenario as Scenario
+from ixmp4.core import Table as Table
 from ixmp4.core import Unit as Unit
 from ixmp4.core import Variable as Variable
-from ixmp4.core import IndexSet as IndexSet
-from ixmp4.core import Scalar as Scalar
-from ixmp4.core import Table as Table
 from ixmp4.core.exceptions import InconsistentIamcType as InconsistentIamcType
 from ixmp4.core.exceptions import IxmpError as IxmpError
 from ixmp4.core.exceptions import NotFound as NotFound
--- a/ixmp4/core/optimization/parameter.py
+++ b/ixmp4/core/optimization/parameter.py
@@ -1,0 +1,123 @@
+from datetime import datetime
+from typing import Any, ClassVar, Iterable
+
+import pandas as pd
+
+from ixmp4.core.base import BaseFacade, BaseModelFacade
+from ixmp4.data.abstract import Docs as DocsModel
+from ixmp4.data.abstract import Parameter as ParameterModel
+from ixmp4.data.abstract import Run
+from ixmp4.data.abstract.optimization import Column
+
+
+class Parameter(BaseModelFacade):
+    _model: ParameterModel
+    NotFound: ClassVar = ParameterModel.NotFound
+    NotUnique: ClassVar = ParameterModel.NotUnique
+
+    @property
+    def id(self) -> int:
+        return self._model.id
+
+    @property
+    def name(self) -> str:
+        return self._model.name
+
+    @property
+    def run_id(self) -> int:
+        return self._model.run__id
+
+    @property
+    def data(self) -> dict[str, Any]:
+        return self._model.data
+
+    def add(self, data: dict[str, Any] | pd.DataFrame) -> None:
+        """Adds data to an existing Parameter."""
+        self.backend.optimization.parameters.add_data(
+            table_id=self._model.id, data=data
+        )
+        self._model.data = self.backend.optimization.parameters.get(
+            run_id=self._model.run__id, name=self._model.name
+        ).data
+
+    @property
+    def constrained_to_indexsets(self) -> list[str]:
+        return [column.indexset.name for column in self._model.columns]
+
+    @property
+    def columns(self) -> list[Column]:
+        return self._model.columns
+
+    @property
+    def created_at(self) -> datetime | None:
+        return self._model.created_at
+
+    @property
+    def created_by(self) -> str | None:
+        return self._model.created_by
+
+    @property
+    def docs(self):
+        try:
+            return self.backend.optimization.parameters.docs.get(self.id).description
+        except DocsModel.NotFound:
+            return None
+
+    @docs.setter
+    def docs(self, description):
+        if description is None:
+            self.backend.optimization.parameters.docs.delete(self.id)
+        else:
+            self.backend.optimization.parameters.docs.set(self.id, description)
+
+    @docs.deleter
+    def docs(self):
+        try:
+            self.backend.optimization.parameters.docs.delete(self.id)
+        # TODO: silently failing
+        except DocsModel.NotFound:
+            return None
+
+    def __str__(self) -> str:
+        return f"<Parameter {self.id} name={self.name}>"
+
+
+class ParameterRepository(BaseFacade):
+    _run: Run
+
+    def __init__(self, _run: Run, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._run = _run
+
+    def create(
+        self,
+        name: str,
+        constrained_to_indexsets: list[str],
+        column_names: list[str] | None = None,
+    ) -> Parameter:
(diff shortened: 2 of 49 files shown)


ixmp4/__init__.py
```diff
@@ -1,16 +1,16 @@
-# flake8: noqa
 import importlib.metadata
 
+from ixmp4.core import IndexSet as IndexSet
```
danielhuppmann, line 3: I'm wondering if we really need to import all of these objects at the top-level? I guess that except for Platform, we will never call them directly, right?

glatterf42: I'm not sure if some downstream packages already use that syntax. We seem to only use it for our tests, where we regularly do `from ixmp4 import Variable`, `from ixmp4 import Model`, or `from ixmp4 import IndexSet`. I don't think this is necessary, but I also don't know what harm it's causing.

ixmp4/core/optimization/parameter.py
```diff
@@ -0,0 +1,123 @@
...
+import pandas as pd
+
+from ixmp4.core.base import BaseFacade, BaseModelFacade
+from ixmp4.data.abstract import Docs as DocsModel
+from ixmp4.data.abstract import Parameter as ParameterModel
+from ixmp4.data.abstract import Run
+from ixmp4.data.abstract.optimization import Column
+
+
+class Parameter(BaseModelFacade):
```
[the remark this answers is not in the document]
glatterf42: They are extremely similar and to some extent, we could have `Parameter` inherit from `Table`. I didn't do that for now because I thought if we then go and update `Table`, it might not be obvious that `Parameter` changes, too. Maybe the cleanest option is to set up a `BaseOptimizationItem` somewhere that contains the common ground between `Table`, `Parameter`, `Variable`, and `Equation`, and have all of them inherit from there.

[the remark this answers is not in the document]
danielhuppmann: Not sure if there is any feature in Table that we would *not* want to be inherited... So better to keep the codebase simple for now and extend later once this becomes relevant.

ixmp4/__init__.py
danielhuppmann: There are no downstream packages yet, so better to keep the top-namespace lean.

ixmp4/core/optimization/parameter.py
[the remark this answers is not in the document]
glatterf42: That's not as straightforward as I thought it would be. Naively trying to extract the maximum number of properties to a common base, I receive
```bash
E   sqlalchemy.exc.NoForeignKeysError: Can't determine the inherit condition between inherited table 'optimization_baseoptimizationitemmodel' and inheriting table 'optimization_table'; tables have no foreign key relationships established.  Please ensure the inheriting table has a foreign key relationship to the inherited table, or provide an 'on clause' using the 'inherit_condition' mapper argument.
```
So it looks like if I set up columns in a common base class, sqlalchemy interprets that as a separate table and inheriting tables need to be linked via ForeignKeys, which is not what we want.

ixmp4/__init__.py
glatterf42: So out of all these imports, which would we want to keep? Only `Platform`?

danielhuppmann: I'd at least not add more imports at this point, and we can do a separate remove-imports-PR later.

danielhuppmann: But yes, I think that only `Platform` is the one that is already used - at least in **pyam**.

glatterf42:
I can't quite get the connection between `Parameter` and `Unit` to work. In theory, we want every `Parameter` to be able to have multiple `Unit`s and every `Unit` (defined for the `Run`) should be usable by multiple `Parameter`s. The `relationship` pattern that most closely resembles this is a [many to many relationship](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#many-to-many). 
These typically work via an association table, i.e. a table in-between `Parameter` and `Unit` that simply holds `Parameter.id` and `Unit.id` pairs, so that each `Parameter` and `Unit` can refer to all pairs that include their own `id`. However, this association table wants to be as efficient as possible, so it half expects and is half recommended to only contain every pair of `(Parameter.id, Unit.id)` just once, and all examples I can find online only show how to set the relationship up that way.
Unfortunately, our needs are more complicated since we want `Parameter.units` to be a list of `Unit`s, which are not necessarily distinct (if I understand correctly, please correct me if I'm wrong). So a `Parameter` may contain several rows of `data` with the same `Unit`, which may even be ordered arbitrarily. Thus, our storage of this `Unit` list needs to preserve order and be able to store the same `Unit` more than once. 
With the current many-to-many setup, I have not been able to make this work. I've also read up on [association objects](https://dev.to/jimgrimes86/flask-sqlalchemy-many-to-many-relationships-association-tables-and-association-objects-3aej) and [association proxy](https://docs.sqlalchemy.org/en/20/orm/extensions/associationproxy.html) in general and I'm not sure we can make it work. We would need to create a purposefully-inefficient table between `Parameter` and `Unit` or figure out a way to do something like this:

1. Receive a list of `Unit`s (via `unit.name`, so we need to `get()` the `Unit`s first)
2. Create entries in the association table with `Parameter.id` and `Unit.id`s
3. Set `Parameter.units` to the `Unit` objects associated with `Parameter`, keeping the order of the input list and potentially `get()`ting the `Unit`s first

The advantages would be that other layers (such as the facade layer) would be able to access `Parameter.units` directly without having to `get()` `Unit`s first; that we could keep as many always-present columns as possible outside of the JSON field `data` (possibly increasing performance); that the relation between `Parameter` and `Unit` would be baked into the DB layer and apparent from there.

The alternative would be to simply keep `units` (and `values`, while we're at it) parts of `data`. This would eliminate both the need to care about order as this is taken care of by JSON and the need to invest more time now to come up with a solution that seems to be pretty non-standard, so would likely be fragile. We could also somewhat bake the relationship into the DB layer through the validator function for `data`. In there, we could `try` to `get()` all distinct `Unit` names once and raise an error if a `Unit` doesn't exist. This way, we would still ensure that all `units` (which we would store as their `unit.name`s) exist in the `Unit` list of the `Run`, as I would understand the [current ixmp docs](https://docs.messageix.org/projects/ixmp/en/stable/data-model.html#pararameter). The only real downside would be that we would return a `Parameter` which doesn't come with the `Unit` objects pre-loaded, so users would have to manually `get()` them if they need them. That sounds acceptable to me.

@danielhuppmann, please let me know what you think.

danielhuppmann:
@glatterf42, I guess that this PR should now be directed to the `main` branch. The files-changed section show quite a number of changes that are already merged.

danielhuppmann:
A comment on the question above by @glatterf42:
>  every `Unit` (defined for the `Run`) 

The units are defined at the `Platform` level (and that should not be changed), I assume that is a typo?

glatterf42:
> @glatterf42, I guess that this PR should now be directed to the `main` branch. The files-changed section show quite a number of changes that are already merged.

The idea was to merge this PR to include/optimization-parameter, which could collect changes to this part of the DB model before it's merged to main. That collection branch should also recognize which commits are already part of main, but maybe this one doesn't since it doesn't know where the collection branch is aiming.

danielhuppmann:
I guess that you have to rebase `optimization\parameter` onto `main` (now and every time other changes are merged there).

glatterf42:
> I guess that you have to rebase `optimization\parameter` onto `main` (now and every time other changes are merged there).

True; at the moment, I'd have to rebase the mixin branch on top of main, the collection branch on top of that, and this branch on top of the new collection branch.

danielhuppmann:
Re the actual, without looking at the code yet (because of the rebase-branch-mixin-mixup):

I think that also our implementation of the "foreign-keys" using index-sets and columns can only be validated at the code level (not the database). Therefore, I think it's ok that we also do that for units, hence storing the columns plus units plus values in `data`.

glatterf42:
Great, thanks :)

glatterf42:
With this clarification, all tests should now work (and that's even more than before), so I'd be in favor of merging this PR and opening another one for the remaining tests and code layers :)

ixmp4/__init__.py
meksor: Another useful one is "Datapoint" because we sometimes need the "Datapoint.Type" enum in user code. 
But i think its time we find a better solution for that anyway...

PR #80, opened 2024-04-16
Support pandas 2.0 and 2.1

danielhuppmann:
This PR implements a work-around for the changed API of pandas 2.2 for `GroupBy.apply()` and adds two tests for older pandas versions.

### DIFF
--- a/ixmp4/data/db/iamc/timeseries/repository.py
+++ b/ixmp4/data/db/iamc/timeseries/repository.py
@@ -128,7 +128,7 @@ class TimeSeriesRepository(
             return df
 
         # ensure compatibility with pandas < 2.2
-        # TODO remove legacy-handing when dropping support for pandas < 2.2
+        # TODO remove legacy-handling when dropping support for pandas < 2.2
         if pd.__version__[0:3] in ["2.0", "2.1"]:
             apply_args = dict()
         else:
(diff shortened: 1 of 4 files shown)


ixmp4/data/db/iamc/timeseries/repository.py
```diff
@@ -127,9 +127,16 @@ def map_measurand(df):
             df["unit__id"] = unit__id
             return df
 
+        # ensure compatibility with pandas < 2.2
+        # TODO remove legacy-handing when dropping support for pandas < 2.2
```
glatterf42, line 131:
```suggestion
        # TODO remove legacy-handling when dropping support for pandas < 2.2
```
Or did you mean `handing`?

glatterf42 (approved):
One possible typo, otherwise looks good to me :)

glatterf42:
I'm just now wondering: should these python 3.10 and 3.11 and different pandas and pyarrow tests really be part of every workflow? Or would it be enough to run our main pytest workflow with one configuration and have a separate legacy workflow for the rest?

ixmp4/data/db/iamc/timeseries/repository.py
danielhuppmann: Thanks for catching that, fixed it.

danielhuppmann:
Yes, we can think of a better test-structure when also introducing a nightly test (only using the latest versions), but let's get the patch-release out so that I can get pyam operational again...

PR #83, opened 2024-04-19
Use recommended dictConfig for logger configuration

glatterf42:
Downstream in https://github.com/iiasa/message-ix-models/issues/173, our most recent release coupled with a recent pyam release led to them using pyam 2.2.0 instead of 1.3.1, which also brought ixmp4 on board. Immediately, their tests for python 3.12 started failing, though I don't know for sure why the tests for other python versions didn't fail. 
Either way, the failure is caused by ixmp4 (and pyam) overwriting the root logger on import, which causes several existing handlers to go missing, one of them being the pytest `caplog` handler. [Thus](https://docs.pytest.org/en/latest/how-to/logging.html#caplog-fixture), tests that were expected to capture certain logging messages didn't receive them anymore.
After reading up on [logging configuration](https://docs.python.org/3/howto/logging.html#configuring-logging), I found that the `FileConfig()` method we currently use has been sort-of-deprecated starting with Python 3.2 -- at least, that's when the more modern `dictConfig()` method was introduced and that's when migration to it has been recommended "in your own time". One major advantage of the `dictConfig()` method is that we don't need to specify anything for the `root` logger, whereas this was required as a key in the `FileConfig` config file. So instead of digging around for a way to set the `root` logger's handlers to the existing ones, I migrated our config to `dictConfig()` :)
The dictionary in question can be generated in any which way you like, but I wanted to keep the structure, i.e. keep them in separate files. First, I considered yaml, but this would add a dependency. Next, I considered toml, but our toml dependency [has last been released in 2020](https://pypi.org/project/toml/#history), and I'm actually surprised it still works. Python gained native toml support via the `tomllib` library with 3.11, so if we want to keep supporting 3.10 (which I think we should), we can't just use that instead. We would still need to specify that for 3.10, the `tomli` dependency is required, which is also different from what we have now but a direct predecessor of the native library. But all this sounds like the topic for a new issue, so I finally opted for json. We can't store comments in json, but apart from that, I'm mostly happy with it.

Locally, the tests are running fine. However, I don't know if we have any tests for confirm that the logging configuration works in the other modes as well. Currently, we have config files for `debug, development, production, server`, but [it looks like](https://github.com/iiasa/ixmp4/blob/25e8532e0cd6327564051628e91df4e6ab442155/ixmp4/conf/settings.py#L26-L28) only the first three are actually allowed and used. Maybe @meksor can confirm: is the `server.conf` used anywhere? If so, we might want to confirm that the `filename` properties of its handlers are evaluated correctly.

Lastly, I'd like to merge these changes soon and publish another patch release to fix the issues downstream. I'll also add a similar PR to pyam.


ixmp4/conf/logging/development.json
```diff
@@ -0,0 +1,21 @@
...
+    "handlers": ["console"]
+  },
+  "handlers":{
+    "console":{
+      "class": "logging.StreamHandler",
+      "level": "NOTSET",
+      "formatter": "generic"
+    }
+  }
+}
```
danielhuppmann, line 21: Add empty line at end of file
```suggestion
}

```

ixmp4/conf/logging/server.json
```diff
@@ -0,0 +1,71 @@
...
+      "filename": "os.getenv('IXMP4_ERROR_LOG')",
+      "mode": "a",
+      "maxBytes": 250000
+    }
+  },
+  "root": { 
+    "level": "NOTSET",
+    "handlers": ["debug","error"]
+  }
+}
```
danielhuppmann, line 71:
```suggestion
}

```

ixmp4/conf/settings.py
```diff
@@ -144,8 +145,10 @@ def configure_logging(self, config: str):
         os.environ.setdefault("IXMP4_DEBUG_LOG", str(debug_file.absolute()))
         os.environ.setdefault("IXMP4_ERROR_LOG", str(error_file.absolute()))
 
-        logging_config = here / f"logging/{config}.conf"
-        logging.config.fileConfig(logging_config, disable_existing_loggers=False)
+        logging_config = here / f"logging/{config}.json"
+        with open(logging_config, "r") as file:
```
danielhuppmann, line 149: Suggest shorten for simplicity
```suggestion
        with open(logging_config) as file:
```

ixmp4/data/api/optimization/indexset.py
```diff
@@ -17,9 +17,13 @@ class IndexSet(base.BaseModel):
...
     id: int
     name: str
-    elements: StrictFloat | StrictInt | list[
-        StrictFloat | StrictInt | StrictStr
-    ] | StrictStr | None
+    elements: (
+        StrictFloat
+        | StrictInt
+        | list[StrictFloat | StrictInt | StrictStr]
+        | StrictStr
```
danielhuppmann, line 24: While we're at it...
```suggestion
        | StrictStr
        | list[StrictFloat | StrictInt | StrictStr]
```

ixmp4/server/rest/optimization/indexset.py
```diff
@@ -21,9 +21,9 @@ class IndexSetInput(BaseModel):
 
 
 class ElementsInput(BaseModel):
-    elements: StrictFloat | StrictInt | list[
-        StrictFloat | StrictInt | StrictStr
-    ] | StrictStr
+    elements: (
+        StrictFloat | StrictInt | list[StrictFloat | StrictInt | StrictStr] | StrictStr
```
danielhuppmann, line 25: As above
```suggestion
        StrictFloat | StrictInt | StrictStr | list[StrictFloat | StrictInt | StrictStr]
```

danielhuppmann (approved):
I have to admit that I still did not fully understand Python logging, but the changes look ok to me and given that @meksor is on annual leave for another week, I think it's ok to merge, do a patch-release, and then discuss later if that is ok for server-side deployment.

PR #85, opened 2024-04-22
Nicer log output on RestAPI connection

danielhuppmann:
A little beautification of the log when connecting to a Platform via the RestAPI.

New output
- the name of the platform
- a warning if the server and client versions are not in line
- the platform notice (if available)

The dsn is now only shown in debug mode.


glatterf42 (approved):
Looks good to me, but I don't often work with the server, so I'm not sure my opinion matters that much here.

meksor (approved):
lgtm

PR #87, opened 2024-05-02
Fix Request dispatching

meksor:
#76 Introduced a problem for programs who themselves are running within something like an [`asyncio.run`](https://docs.python.org/3/library/asyncio-runner.html#asyncio.run) call. 
Since this is the case for any jupyter notebook this is not very convenient. 
I switched the approach to use a `ThreadPoolExecutor` instead and it seems to work, please verify @danielhuppmann.

### DIFF
--- a/ixmp4/conf/settings.py
+++ b/ixmp4/conf/settings.py
@@ -36,7 +36,7 @@ class Settings(BaseSettings):
     max_page_size: int = 10_000
     default_page_size: int = 5_000
     client_default_upload_chunk_size: int = 10_000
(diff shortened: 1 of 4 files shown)


meksor:
Tests failing b/c of manager database outage...

danielhuppmann:
Thanks @meksor - works for me in a Jupyter notebook now (also with requesting all data from ssp-extensions, 1.5m rows), but two tests are failing because of shut-down future...

meksor:
fixed!

danielhuppmann (approved):
Works in my notebook again, many thanks!

PR #91, opened 2024-05-28
Better handling of `dimensionless` unit

danielhuppmann:
This PR removes the special handling of "dimensionless" units, previously introduced because ORACLE had problems with storing empty strings.

Creating units "dimensionless" or multiple whitespaces is still not allowed, users should use an empty string to indicate dimensionless units.

### DIFF
--- a/ixmp4/core/iamc/data.py
+++ b/ixmp4/core/iamc/data.py
@@ -13,15 +13,6 @@ from ..utils import substitute_type
 from .variable import VariableRepository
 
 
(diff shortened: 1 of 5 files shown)


danielhuppmann:
Per discussion at the weekly standup just now, we decided to not implement an automated migration, because there are only two ixmp4 platforms currently in production, and we can update those manually.

meksor (approved):
lgtm

PR #92, opened 2024-06-18
Fix Multi-Process Logging

meksor:
The migration to kubernetes has revealed a bug in the use of `RotatingFileHandler`. I am guessing since the write delays on nfs are higher than on the old file system, sometimes when a file is supposed to be rotated another process tries to write to it at the same time. Apparently `RotatingFileHandler` gets a new file descriptor when rotating, leaving all file descriptors except in the Handler that rotated the file, invalid. 
I've disabled file logging completely as a quick fix (kubernetes will keep the stdout logs a little while for us) and turned on console logging for everything. Hopefully we can switch over to a log aggregation service soon so we don't need file logging at all.
I also noticed that we print some passwords to the logs, i've fixed that here as well.

### DIFF
--- a/ixmp4/conf/logging/server.json
+++ b/ixmp4/conf/logging/server.json
@@ -9,24 +9,22 @@
   },
   "loggers": {
     "fastapi": {
-      "level": "NOTSET",
-      "handlers": ["console"]
+      "level": "NOTSET"
     },
     "httpx": {
-      "level": "NOTSET",
-      "handlers": ["console"]
+      "level": "NOTSET"
     },
     "sqlalchemy": {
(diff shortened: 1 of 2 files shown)


glatterf42 (approved):
Looks good to me :)

Should we immediately open a tracking issue for switching to a log aggregation service or will we remember ourselves?

meksor:
Youre right! Probably best to do this in scse-infra or internally somewhere else since it just requires changes to the deployment config, not ixmp4.

meksor:
https://github.com/iiasa/scse-infra/issues/63

Issue #93, opened 2024-06-24
pydantic error


PBL-markroelfsema:
When I try to install (via Conda) the ixmp4 package, I get the following error

```bash
Traceback (most recent call last):
  File "X:\user\roelfsemam\environments\glasgow_init_env\Scripts\ixmp4-script.py", line 5, in <module>
    from ixmp4.__main__ import app
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\__init__.py", line 4, in <module>
    from ixmp4.core import Model as Model
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\core\__init__.py", line 2, in <module>
    from .iamc.variable import Variable as Variable
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\core\iamc\__init__.py", line 3, in <module>
    from .data import RunIamcData, PlatformIamcData
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\core\iamc\data.py", line 9, in <module>
    from ixmp4.data.backend import Backend
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\data\backend\__init__.py", line 3, in <module>
    from .api import RestBackend, RestTestBackend
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\data\backend\api.py", line 8, in <module>
    from ixmp4.conf import settings
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\conf\__init__.py", line 3, in <module>
    from ixmp4.conf.settings import Settings
  File "C:\Users\roelfsemam\AppData\Roaming\Python\Python311\site-packages\ixmp4\conf\settings.py", line 8, in <module>
    from pydantic import Field, HttpUrl, field_validator
ImportError: cannot import name 'field_validator' from 'pydantic' (X:\user\roelfsemam\environments\glasgow_init_env\Lib\site-packages\pydantic\__init__.cp311-win_amd64.pyd)
```

On StackOverlfow they seem to suggest that the call to pydantic has changed to 
from pydantic.functional_validators import field_validator
https://stackoverflow.com/questions/77172420/importerror-cannot-import-name-field-validator-from-pydantic

PBL-markroelfsema:
[...]

PR #95, opened 2024-06-28
Update dependency lock file 2024/06

glatterf42:
Looking at #93, I noticed that I was using an outdated version of pydantic locally. So this PR updates the dependency lock file that was automatically created by running `poetry update` locally. No additional changes are required :)
Most notably, everything works even without warnings after bumping to `numpy 2.0.0`.


glatterf42:
Well, so much for the 'no changes required' part. It seems that already with [`poetry 1.7.0`](https://python-poetry.org/blog/announcing-poetry-1.7.0/), 'setuptools' was no longer installed per default. I don't know why it was removed _now_ on my local system, but if we won't to keep it around (which is required by our docs-build-process), we should include it as an explicit dependency.

glatterf42:
`pandas` does not specify upper bounds for `numpy`, even though some fairly recent versions of `pandas` don't support `numpy 2.x`. We test these versions by pinning `pandas`, and we only pin `pandas` after initially installing everything as normal. This installs `numpy 2.x`, but then only `pandas` is downgraded, leading to a binary incompatibility with `numpy`. 
So at the moment, this "legacy pandas step" is now also pinning `numpy 1.26.4`, which is the latest version before `2.0.0`. Please let me know if this is acceptable.

Also, add color support to pytest's GHA output because

> [!CAUTION]
> More

> [!WARNING]
> color

> [!TIP]
> is

> [!NOTE]
> always

> [!IMPORTANT]
> good.

(Though surprisingly little supported by Github.)

meksor:
Fine from my side! Lgtm ty

meksor (approved):

PR #96, opened 2024-07-01
DRY optimization core layer

glatterf42:
For #94, @danielhuppmann asked in the `core` layer: "Why does `Parameter` not inherit from `Table`"? And sure enough, these optimization items share a lot of common functionality. So before moving on to `Variable` and `Equation`, for which this will be the case even more so, I wanted to take another look at this.

@meksor, I expect you'll be the one to take a look at this. Sorry, this description might be a bit lengthy.

Initially, I interpreted the question as for the DB layer, and how to handle inheritance (or rather _Composition_) there has been recently discussed in #82. Instead, this PR focuses on the core layer, where we also have two classes per optimization item that could potentially benefit from higher reusability (to follow the DRY principle): e.g., the `Table` (mostly a collection of data attributes) and the `TableRespository` (functions to handle `Table`s).
For the `Table`s themselves (and similar), we are currently using Python's [property](https://docs.python.org/3/library/functions.html#property) to make data attributes accessible in a controlled way. Unfortunately, these `property`s don't mix well with inheritance. [This article](https://gist.github.com/Susensio/979259559e2bebcd0273f1a95d7c1e79) describes how to still make it work for the `getter` part, but since we would still have to repeat the same number of lines in the data files, I'm not sure it's worth the effort here. What we stand to gain is a separation of functionality and placeholder code, it seems to me. So when we want to make changes to how e.g. `data` is returned from an optimization model, we would just need to change one place and all others would inherit these changes. However, code repetition would not be decreased and understanding the models might become more complicated, so I've put this on hold for now.
For `TableRepository`, there is luckily a quite straightforward way: we can employ the same approach as in the DB layer and create an `OptimizationBaseRepository` in the core layer that can be extended with `Creator`, `Retriever`, etc as needed. This seems to work well and every optimization item repo then just needs to clarify which `_model_type` and `_backend_repository` it is interested in. 
For better understanding, I wanted to type hint the `_backend_repository` in the `OptimizationBaseRepository`. However, we didn't already have an appropriate object to do so. That's why I have created a `BackendBaseRepository` in `abstract.optimization.base`, which simply announces how the actual `TableRepository` in the DB layer is going to look. I'm not sure if this is the correct location; I've usually worked with abstract models to type hint API/DB layer functionality, but I also don't know where else it would live.

This is the point where I'm not sure I'm naming things correctly anymore. I started from `Table`, which already has `Table.data`, which is arguably the largest part of shared functionality, at least in the DB layer, and which is missing from `IndexSet` and `Scalar`. So these two items could not use the `OptimizationBaseRepository` as it is now, which seems like a misnomer. Maybe the data functionality needs to be further separated (allowing us to DRY even more parts of this layer) or maybe we find another name?
The same applies for `BackendBaseRepository`.

-----

> [!NOTE]
> For the curious: there's an [issue on CPython](https://github.com/python/cpython/issues/59170) about `property`s and inheritance from 2012.

### DIFF
--- a/ixmp4/core/optimization/base.py
+++ b/ixmp4/core/optimization/base.py
@@ -49,7 +49,7 @@ class Lister(OptimizationBaseRepository[OptimizationModelType], abstract.Lister)
 
 class Tabulator(OptimizationBaseRepository[OptimizationModelType], abstract.Tabulator):
     def tabulate(self, name: str | None = None) -> pd.DataFrame:
-        return self._backend_repository.tabulate(name=name)
+        return self._backend_repository.tabulate(run_id=self._run.id, name=name)
 
 
 class Enumerator(
--- a/ixmp4/core/optimization/base.py
+++ b/ixmp4/core/optimization/base.py
@@ -1,39 +1,9 @@
-# from typing import Any, ClassVar, Generic, TypeVar
 from typing import Generic, TypeVar
 
 import pandas as pd
 
 from ixmp4.core.base import BaseFacade, BaseModelFacade
-
-# from ixmp4.core.exceptions import IxmpError
 from ixmp4.data import abstract
-
-# class OptimizationBaseModel(BaseModelFacade):
-#     _model: abstract.OptimizationBaseModel
-#     NotFound: ClassVar[type[IxmpError]]
-#     NotUnique: ClassVar[type[IxmpError]]
-
-#     @property
-#     def id(self) -> int:
-#         return self._model.id
-
-#     @property
-#     def name(self) -> str:
-#         return self._model.name
-
-#     @property
-#     def run_id(self) -> int:
-#         return self._model.run__id
-
-#     @property
-#     def data(self) -> dict[str, Any]:
-#         return self._model.data
-
-#     # TODO Stopping here since I'm not convinced of the usefulness. The only benefit I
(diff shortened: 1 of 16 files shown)


meksor:
OK so i think the currently proposed changes (removing the commented base model) are good. I also think the rest of the facade layer could also benefit from something like this. Lets figure the `@property` part out later, for now just explicitly having them in every leaf class is okay imo. 
Since we are now not depending on sqlalchemys session-synchronization anyway (changing something in a model with id=1 will change it in all models referencing that row) it would also be okay to just reassign the properties (maybe automatically) to the facade class from the api or db model, thereby completely removing the whole property thing for most stuff.

meksor (approved):
noticed that i never approved this

glatterf42:
Please note: with [d964068](https://github.com/iiasa/ixmp4/pull/96/commits/d96406804fc508c723b15e953e16e450544a5a49), this PR is fine to be merged to `main`. All tests are passing and it was approved.

However, it adds a lot of untyped code, so I will rebase it onto #129's branch and resolve all merge conflicts/add type hints here, too. In fact, I'll even use #138's branch since that makes even more relevant changes. Once that works, we can merge it; should it turn out to be too complicated, we can still revert to the commit mentioned above.

PR #99, opened 2024-07-05
All we need to do is listen.

meksor:
Hi
I've replaced the current audit information setup (calling `set_creation_info` when creating an object) with a listener system in anticipation for the versioning change.

I've added the class `SqlaEventHandler` to be instantiated in each database backend. This handler listens to three events explained in the docstrings. 

If the event targets inherit from the new mixins (by fridolin, moved and refactored by me here), special data will be injected into the queries that serves to set/update the relevant columns without an extra query dispatch. This is documented in the mixin docstrings.

The following test was added:
```py
def test_audit_info(self, test_mp, request):
    test_mp = request.getfixturevalue(test_mp)
    run = test_mp.backend.runs.create("Model", "Scenario")
    test_mp.backend.runs.set_as_default_version(run.id)
    test_mp.backend.runs.create("Model", "Scenario")

    runs = test_mp.backend.runs.tabulate(default_only=False)
    assert (runs["created_by"] == "@unknown").all()
    # was updated by set_as_default_version
    assert runs["updated_by"][0] == "@unknown"
    assert runs["updated_by"][1] is None
```
As far as I can see this is the only place where the current rules apply. `Run` is the only model that can be updated directly (so i added `HasUpdateInfo` to its parents).

**Note** that this does not include the functionality to automatically update the `UpdateInfo` when related objects are changed.
For example when using `Run.iamc.add` or `Run.iamc.remove`.
This is on purpose as I think it will make these PRs easier to review. My plan is to implement some sort of "ownership" system that can be used for update information, orphan deletion and permission checking all at the same time. A short summary could be that we want to allow rows to own other rows along foreign keys. For example a Run owns multiple TimeSeries which in turn own multiple DataPoints. More to follow in a seperate PR.

**More:**
 - Removed `async_client` from api backend as it was emitting deprc. errors and is no longer used.

### DIFF
--- a/ixmp4/data/db/mixins.py
+++ b/ixmp4/data/db/mixins.py
@@ -34,7 +34,7 @@ class HasCreationInfo:
 
 
 class HasUpdateInfo(HasCreationInfo):
-    """Mixin for adding creation audit info to a model.
+    """Mixin for adding update and creation audit info to a model.
     The added fields `updated_at`, `updated_by` are automatically set
     by ixmp4's sqlalchemy event handlers for 'before_update' and 'do_orm_execute'
     (in addition to the inherited fields from `HasCreationInfo`).


ixmp4/data/db/mixins.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_timestamp():
+        return datetime.now(tz=timezone.utc)
+
+    def set_creation_info(self, auth_context: "AuthorizationContext | None") -> None:
+        self.created_at = self.get_timestamp()
+        self.created_by = self.get_username(auth_context)
+
+
+class HasUpdateInfo(HasCreationInfo):
+    """Mixin for adding creation audit info to a model.
```
glatterf42, line 37:
```suggestion
    """Mixin for adding update audit info to a model.
```
Is this description correct?

glatterf42 (approved):
Looks good to me, thanks :)
Just one question about wording in-line :)

ixmp4/data/db/mixins.py
meksor: Probably best to both mention creation and update info, but this is a mistake, yes!`

```suggestion
    """Mixin for adding update and creation audit info to a model.
```

PR #100, opened 2024-07-08
Specify ruff check to avoid deprecation

glatterf42:
Just ran some migration locally and was informed that `ruff <path>` is now deprecated, we should use `ruff check <path>` instead. This PR updates our alembic post write hook accordingly.


meksor (approved):
lgtm

PR #101, opened 2024-07-10
Include the linopy version of the transport tutorial

glatterf42:
> [!NOTE]
> This PR looks huge, but it actually isn't. Please focus on `py_transport.ipynb` and `linopy_model.py` in `transport/tutorial`.

The reason this PR looks so huge is because all my others PRs are still waiting for reviews but they are actually needed to enable the transport tutorial. Thus, I have `git cherry-pick`ed the required commits from each one of them, increasing the apparent amount of changes in here. However, the actually important changes are only about 200 lines long, so please bare with me.

First of, this version is using linopy because that was more readily available than learning how to use gamsapi (which I will do next, regardless) and because it might help to then abstract some `ixmp4-solver` package or some such so that it doesn't lean too much on the gams formulation. Linopy support a wide range of solvers, of which I have employed the `highs` solver. I have installed the full [highs](https://github.com/ERGO-Code/HiGHS) software locally and added the `highspy` package as a dependency. I'm not sure if just the python package is enough; please install the full software in that case (all check should still pass because ...

- [ ] ... we don't have any test for the tutorial yet)

Linopy also supports gurobi, which unfortunately does not work with numpy 2.0 yet. Hence linopy's pin of numpy 2.0. Since we do want to support recent numpy versions, though, I have created a branch on my linopy fork that removes the numpy pin and added this branch as a dependency here. Please note that removing this pin does break linopy's test suite locally, namely the test relying on gurobipy. Please also note that this branch may go out of date since linopy is actively developed and I don't know how frequent I can update my fork.

This PR contains two changes to the data validation functions for `Equation`s and `Variable`s because we actually only need to validate data if there is data to validate. The corresponding commit should be `cherry-pick`ed to the other branches so that they can be merged and this PR can be rebased, or it can be included here.

Similarly, this PR contains a commit introducing all missing DB migrations, which is a lot. These might still change due to the pending reviews on other PRs, so this commit is marked as `TEMPORARY`.

Finally, the actual tutorial code is included in the files mentioned in the beginning. 
GAMS would need a model file, if I understand correctly, to record all parameters, variables, and equations properly. Linopy would not need that, but I figured I could try to abstract data from functionality and so have put the functionality almost completely into `linopy_model.py`. Doing so means that the tutorial itself only needs a few lines of code to work with linopy, which is probably not enough to warrant a new `ixmp4-solver` package.
But please share your thoughts on this :)


tutorial/transport/py_transport.ipynb
```diff
@@ -383,7 +365,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "scen.solve(model='dantzig')"
+    "from tutorial.transport.linopy_model import create_dantzig_model, store_dantzig_solution\n",
+    "\n",
+    "m = create_dantzig_model(i=i, j=j, a=a, b=b, d=d, f=f)\n",
```
danielhuppmann, line 370: The Dantzig model code knows which sets and parameters are needed so this can all be read from the `run` directly...
```suggestion
    "m = create_dantzig_model(run)\n",
```

tutorial/transport/py_transport.ipynb
```diff
@@ -383,7 +365,12 @@
...
+    "m.solve(\"highs\")\n",
+    "\n",
+    "store_dantzig_solution(model=m, z=z, x=x, demand=demand, supply=supply)"
```
danielhuppmann, line 373: Suggest to rename that method, and as before, the Dantzig model code knows which variables and equations to read.
```suggestion
    "read_dantzig_solution(run, model=m)"
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
+# m = linopy.Model()
```
danielhuppmann, line 1: Suggest to rename this file to "dantzig_model_linopy.py" (because we may have other models later).

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+    return pd.Series(data=parameter.values, index=index, name=name)
+
+
+def create_dantzig_model(
+    i: IndexSet,
+    j: IndexSet,
+    a: Parameter,
+    b: Parameter,
+    d: Parameter,
+    f: Scalar,
```
danielhuppmann, line 51: Per my other comment in the notebook, this should be streamlined to only take the `run`. Looking ahead to a MESSAGE use case, we don't want to list all sets and parameters when calling something like
```
message_ix.run(Run)
```
(where `run()` is a shorthand for writing to gdx, calling GAMS to solve, reading the solution).
```suggestion
    run: Run,
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+) -> linopy.Model:
+    m = linopy.Model()
+    i_set = create_set(indexset=i, name="Canning Plants")
```
danielhuppmann, line 54: I guess that you don't need the `create_set()` utility.... And I'm not sure why you give long names instead of the short ones used in the mathematical notation (why "Canning Plants", not "i").
```suggestion
    i_set = pd.Index(
       run.optimization.indexsets.get("i").elements, 
       name="i",
    )
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+
+    c = d_parameter * f_scalar / 1000
+    c.name = "transport cost in thousands of dollars per case"
+
+    x = m.add_variables(
+        lower=0.0, coords=[i_set, j_set], name="Shipment quantities in cases"
+    )
+
+    con = x.sum(dim="Markets") <= a_parameter
+    m.add_constraints(con, name="Observe supply limit at plant i")
```
danielhuppmann, line 75: Maybe more pythonic?
```suggestion
    m.add_constraints(
        x.sum(dim="Markets") <= a_parameter,
        name="Observe supply limit at plant i"
    )
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+    m.add_constraints(con, name="Satisfy demand at market j")
+
+    obj = c.to_xarray() * x
+    m.add_objective(obj)
+
+    return m
+
+
+def store_dantzig_solution(
+    model: linopy.Model, z: Variable, x: Variable, demand: Equation, supply: Equation
```
danielhuppmann, line 87:
```suggestion
    model: linopy.Model, run: Run
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+    # Can't add units if this column was not declared above. Better stored as Scalar
+    # maybe?
+    z.add(data={"levels": [model.objective.value], "marginals": [-0.0]})
+
+    # Handle shipment quantities
+    x_data: pd.DataFrame = model.solution.to_dataframe()
+    x_data.reset_index(inplace=True)
+    x_data.rename(
+        columns={
+            "Shipment quantities in cases": "levels",
```
danielhuppmann, line 100: A bit confused why "levels" is plural...

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+            "Canning Plants": "i",
+            "Markets": "j",
+        },
+        inplace=True,
+    )
+    # x_data["units"] = "cases"
+    # TODO Again setting fake marginals until they are optional for variables
+    x_data["marginals"] = -0.0
+    x.add(data=x_data)
```
danielhuppmann, line 109:
```suggestion
    run.optimization.variables.get("x").add(data=x_data)
```

tutorial/transport/linopy_model.py
```diff
@@ -0,0 +1,126 @@
...
+    f_scalar = f.value
+
+    c = d_parameter * f_scalar / 1000
+    c.name = "transport cost in thousands of dollars per case"
+
+    x = m.add_variables(
+        lower=0.0, coords=[i_set, j_set], name="Shipment quantities in cases"
+    )
+
+    con = x.sum(dim="Markets") <= a_parameter
```
danielhuppmann, line 74: This would be closer to the mathematical formulation.
```suggestion
    con = x.sum(dim="j") <= a_parameter
```

danielhuppmann (commented):
Thanks @glatterf42, this is coming together nicely! I made a few suggestions for streamlining, hope that makes sense...

tutorial/transport/py_transport.ipynb
```diff
@@ -223,22 +207,36 @@
...
+    "except Unit.NotFound:\n",
+    "    cases = mp.units.create(\"cases\")\n",
+    "\n",
+    "# capacity of plant i in cases\n",
+    "# add parameter data as a dict\n",
+    "a = run.optimization.parameters.create(name=\"a\", constrained_to_indexsets=[\"i\"])\n",
+    "a_data = {\n",
+    "    \"i\": [\"seattle\", \"san-diego\"],\n",
+    "    \"values\": [350, 600],\n",
+    "    \"units\": [cases.name, cases.name],\n",
```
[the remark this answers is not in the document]
glatterf42: Yes, because they describes lists of values and units. The alternative would be to have a column called `value`, which actually represents multiple values.

A very similar issue could be discussed for `levels` and `marginals` for `Variable` and `Equation`. I've applied the same reasoning there.

tutorial/transport/py_transport.ipynb
glatterf42: I thought about that and I'm not sure I agree. Passing only a `Run` means we have to document the requirements for this `Run`: it needs to have these `IndexSet`s, `Parameter`s, etc. The way it is now, we only explicitly pass in the data we need, or maybe even still too much. In practice, it might suffice to just pass in `i.elements` and `i.name` etc.

tutorial/transport/py_transport.ipynb
glatterf42: That was how I originally named it and then I thought: the more important part is probably storing it with the `Run`. But I can change back of course ;)

danielhuppmann: Yes, please.

tutorial/transport/py_transport.ipynb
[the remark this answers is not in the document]
danielhuppmann: I see that you applied this reasoning consistently, but it is a departure from the **ixmp** API. It is also quite clear that the column "value" has all values, so in the interest of least disruption for existing users, I'd keep the ixmp convention.

tutorial/transport/linopy_model.py
glatterf42: The pyomo and linopy tutorials use long names and I'm usually a fan of that as well. Because on the flipside: why would you use `i` instead of `Canning Plants`? Maybe you can save a little bit of time, but you create more confusion because people new to the problem then constantly have to check: what was `i` again? And we could even do something like 
```python
i_name = "Canning Plants"
i_set = create_set(indexset=i, name=i_name)
```
so that we have to completely type it only once and still have clarity in the variables and results: what did this actually mean?

tutorial/transport/linopy_model.py
glatterf42: Originally, linopy prefers the way I did it because you can then do `print(con)` and it shows you the exact form of the constraint you're about to add. You can use that to verify it, but we don't have to keep that, of course.

tutorial/transport/py_transport.ipynb
danielhuppmann: But  think about the signature for creating a message_ix model, which has dozens of sets and parameters. This will become extremely painful.

In practice, the "model" knows which sets and parameters are required, so if a **Run** is passed that doesn't have a required item, the model (i.e., the model-creation-function) will just raise an error.

tutorial/transport/linopy_model.py
danielhuppmann: I understand that this can be helpful when developing the model, but in your script, you overwrite `con` immediately anyway, so there is no benefit for the more verbose approach in a well-developed model-creation script.

tutorial/transport/py_transport.ipynb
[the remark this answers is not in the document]
glatterf42: Okay, I will change to `value` and `unit` here. For consistency, we should then also probably change to `indexset.element`, `variable.level` etc, which I don't want to do here because the purpose of this PR should be to bring in the tutorial, not to change unmerged DB content.

tutorial/transport/py_transport.ipynb
glatterf42: Theoretically, the signature for something like `create_message_model()` could be different from `create_dantzig_model()`, but fine, I'll just pass a run.

glatterf42: Please note, though, that passing a run requires additional DB calls to `get()` the indexsets etc, which likely slows down runtime. I doubt that it will notably affect the transport tutorial, though.

tutorial/transport/py_transport.ipynb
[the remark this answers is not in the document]
danielhuppmann: Ok to do this in other PRs, sorry that I didn’t catch it earlier. But we don’t want to ask colleagues to update all their scripts for MESSAGE data pre- or postprocessing. Maybe the item attributes can stay as is, but the tabular representation should be consistent with **ixmp**.

tutorial/transport/linopy_model.py
glatterf42: Done, but please note that we now have
```python
from tutorial.transport.dantzig_model_linopy import (
    create_dantzig_model,
    read_dantzig_solution,
)
```
where I read `transport` as equivalent to `dantzig` and thus see a lot of repetition.

danielhuppmann: Don't think that this is too bad...

tutorial/transport/linopy_model.py
glatterf42: Huh, I was just tinkering with this, but now I'm thinking: if this is supposed to be a tutorial, the way we have it now might do that best: showing people how to work with linopy and ixmp4. Because it might not be clear, that the dimension names of the linopy objects are allowed to differ or where exactly these differences can occur. With the current setup, we would demonstrate that.
So even without harmonizing everything to read either `i` or `canning_plants`, I think the current setup is useful.

BTW: I have removed the helper function as suggested here.

tutorial/transport/py_transport.ipynb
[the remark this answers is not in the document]
glatterf42: Turns out: I can't even do that for `parameter.values` and `parameter.units` as these column names are also required (explicitly in plural form). But let's create a new issue from this comment and keep this in mind when reviewing the other PRs. I would suggest merging the other PRs without fixing this (requires less rework of this tutorial) and then have a separate clean up PR for this issue fixing at the DB level and the tutorial.

To be clear: at the core level, you want the `parameter.values` etc to stay plural form?

tutorial/transport/linopy_model.py
danielhuppmann: There is no right or wrong between short names (close to the math) or long names (readability), either in the notebook or the model-code, as long as it’s understandable and only the run is passed as argument, not the individual items.

glatterf42:
Good news first: we now have tests for the tutorial :)
Bad news: 

- I'm not entirely sure why one of them isn't working, commented that one out for now (not too familiar with xarray yet, I'd prefer to use pandas).
- I had to duplicate the model file (more or less) because I didn't know where to put it in relation to existing files. This might be fine if we actually migrate all tutorial/solver stuff into their own package, otherwise, @meksor, please let me know where to put the files.

Still wanted to push these changes before going on leave.

tutorial/transport/py_transport.ipynb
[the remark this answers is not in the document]
glatterf42: Continued in #102.

danielhuppmann:
> @danielhuppmann we might want to include the optimization.solve functionality only as a subpackage or similar. Do you have a preferred way of integrating the functionality now that it could be included (after some clean up)?

If I understand this correctly, the `solve` method is part of linopy anyway.

Looking at the PR quickly, it seems like the current approach is - makes sense to me.
```python
m = linopy.Model()
# transfer sets and parameters to `m`
m.solve()
# extract solution and save as Variables
```

And for more elaborate model-frameworks like MESSAGEix, this can be part of the message_ix package, like
```python
run = message_ix.Scenario()  # this implicitly creates the sets, parameters, variables
# populate the sets and parameters with data
run.solve()  # which transfers sets and parameters, calls `solve()`, and extracts the solution and saves as Variables
# get variables from `run` to analyze the results
```

PR #104, opened 2024-07-31
Add audit info migration

meksor:
adds a migration for #99


glatterf42 (approved):
Thanks for adding this DB migration! Didn't run into problems yet, but let's see if it affects the open DB-layer PRs. 

Looks good to me once tests pass.

PR #106, opened 2024-08-02
Use opearator functions instead of our own where possible

glatterf42:
Today I learned that Python comes with the [operator module](https://docs.python.org/3/library/operator.html). This provides various operators as functions, which made me think of `db/filters.py`, where we define several functions again like `exact`, which is equal to `operator.eq`. 
This PR reduces possible errors and maintenance for us by replacing our functions with those from `operator`. Please note that more might be possible; there is e.g. [`operator.contains`](https://docs.python.org/3/library/operator.html#operator.contains), but it's not a drop-in replacement for `in_` (I tried, some tests failed) and I didn't want to spend too much time on this, so for now, it's not included here.

### DIFF
--- a/ixmp4/db/filters.py
+++ b/ixmp4/db/filters.py
@@ -1,3 +1,4 @@
+import operator
 from types import UnionType
 from typing import Any, ClassVar, Optional, Union, get_args, get_origin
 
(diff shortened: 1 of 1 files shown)


meksor (approved):
lgtm

PR #107, opened 2024-08-02
Fix `list()` and `tabulate()` for existing optimization items with specific run-request

glatterf42:
Also while working on the GAMS transport tutorial, I discovered that all optimization items had inconsistent behavior in the facade layer: `run.optimization.scalar.tabulate()` would list all scalars of all runs, while `...list()` would list only those of the run the scalars belong to. The latter is clearly desired, so this PR fixes the inconsistency and adds tests to confirm. 

Since `Parameter`, `Equation`, and `Variable` have not yet landed on main, I'll include corresponding changes in their pending PRs.

### DIFF
--- a/tests/core/test_indexset.py
+++ b/tests/core/test_indexset.py
@@ -51,6 +51,7 @@ class TestCoreIndexSet:
     def test_get_indexset(self, test_mp, request):
         test_mp = request.getfixturevalue(test_mp)
         run = test_mp.runs.create("Model", "Scenario")
+        run.set_as_default()
         _ = run.optimization.indexsets.create("IndexSet 1")
         indexset = run.optimization.indexsets.get("IndexSet 1")
         assert indexset.id == 1
@@ -90,8 +91,6 @@ class TestCoreIndexSet:
     def test_list_indexsets(self, test_mp, request):
         test_mp = request.getfixturevalue(test_mp)
         run = test_mp.runs.create("Model", "Scenario")
-        # Per default, list() lists only `default` version runs:
-        run.set_as_default()
         indexset_1 = run.optimization.indexsets.create("Indexset 1")
         indexset_2 = run.optimization.indexsets.create("Indexset 2")
         expected_ids = [indexset_1.id, indexset_2.id]
@@ -106,11 +105,17 @@ class TestCoreIndexSet:
         ]
         assert not (set(expected_id) ^ set(list_id))
 
+        # Test that only indexsets belonging to this Run are listed
+        run_2 = test_mp.runs.create("Model", "Scenario")
+        indexset_3 = run_2.optimization.indexsets.create("Indexset 1")
+        indexset_4 = run_2.optimization.indexsets.create("Indexset 2")
+        expected_ids = [indexset_3.id, indexset_4.id]
+        list_ids = [indexset.id for indexset in run_2.optimization.indexsets.list()]
+        assert not (set(expected_ids) ^ set(list_ids))
+
     def test_tabulate_indexsets(self, test_mp, request):
         test_mp = request.getfixturevalue(test_mp)
         run = test_mp.runs.create("Model", "Scenario")
-        # Per default, tabulate() lists only `default` version runs:
-        run.set_as_default()
         indexset_1 = run.optimization.indexsets.create("Indexset 1")
         indexset_2 = run.optimization.indexsets.create("Indexset 2")
         expected = df_from_list(indexsets=[indexset_1, indexset_2])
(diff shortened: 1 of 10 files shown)


tests/core/test_indexset.py
```diff
@@ -106,11 +105,17 @@ def test_list_indexsets(self, test_mp, request):
         ]
         assert not (set(expected_id) ^ set(list_id))
 
+        # Test that only indexsets belonging to this Run are listed
```
danielhuppmann, line 108: Wouldn't it be much less verbose to add the following after creating `indexset_2`?

```python
# create an indexset in another run to make sure that the correct indexsets are listed
test_mp.runs.create("Model", "Other Scenario").optimization.indexsets.create("Other Indexset")
```

tests/core/test_indexset.py
```diff
@@ -123,6 +128,16 @@ def test_tabulate_indexsets(self, test_mp, request):
         result = run.optimization.indexsets.tabulate(name="Indexset 2")
         pdt.assert_frame_equal(expected, result)
 
+        # Test that only IndexSets belonging to this Run are tabulated
```
danielhuppmann, line 131: Same comment as above, better to create another indexset earlier in this test.

tests/core/test_indexset.py
```diff
@@ -90,8 +91,6 @@ def test_add_elements(self, test_mp, request):
     def test_list_indexsets(self, test_mp, request):
         test_mp = request.getfixturevalue(test_mp)
         run = test_mp.runs.create("Model", "Scenario")
-        # Per default, list() lists only `default` version runs:
-        run.set_as_default()
         indexset_1 = run.optimization.indexsets.create("Indexset 1")
```
danielhuppmann, line 94: Might be useful to put this index set-creation into a utility function to avoid duplication of code?

danielhuppmann (commented):
Thanks @glatterf42, good catch! I made a few comments on streamlining the tests for the indexsets, does also apply to the other types...

tests/core/test_indexset.py
glatterf42: Very good catch, I'll rework the tests here and in the other PRs.

tests/core/test_indexset.py
glatterf42: The idea sounds good and I implemented it, but dealing with the new function actually takes more lines than before in some cases. In particular, the function uses the backend layer to maximize speed, but the core layers sometimes need the core objects returned instead of abstract/backend ones, so they need to be converted.

danielhuppmann: Ok, thanks for trying

glatterf42: I think the solution we now have is actually fine and I would also bring that to the other branches. Or do you want me to revert that change?

tests/utils.py
```diff
@@ -198,3 +200,16 @@ def create_iamc_query_test_data(test_mp):
     run2.meta = {"run": 2, "test": "string", "bool": False}
 
     return [r1, r2, r3], units
+
+
+def create_indexsets_for_run(
+    platform: Platform, run_id: int, amount: int = 2, offset: int = 0
+) -> tuple[IndexSet, ...]:
+    """Create `amount` indexsets called `Indexset n` for `run` (n in (offset,
+    offset+amount])."""
```
danielhuppmann, line 209: To keep it in one line...
```suggestion
    """Create `amount` indexsets called `Indexset n` for a `run`."""
```

tests/core/test_indexset.py
danielhuppmann: Sorry, I misunderstood your response, it thought that you decided against actually using this because it used more lines of code...

tests/utils.py
```diff
@@ -198,3 +200,16 @@ def create_iamc_query_test_data(test_mp):
     run2.meta = {"run": 2, "test": "string", "bool": False}
 
     return [r1, r2, r3], units
+
+
+def create_indexsets_for_run(
+    platform: Platform, run_id: int, amount: int = 2, offset: int = 0
```
danielhuppmann, line 206: You could set offset to 1 as default and remove the "+1" on line 212.

danielhuppmann (approved):
Two minor suggestions, then good to go as far as can tell!

tests/utils.py
glatterf42: That's what I had originally and then I thought I should clarify what `amount` and `offset` are, too. But maybe we can consider them self-explanatory.

danielhuppmann: I think there's beauty in one-line docstrings, and the function is really short enough that it's easy enough to understand.

glatterf42: Agreed, changed it in all place where this function is introduced :)

tests/utils.py
glatterf42: Good catch, I did not work with full concentration, it seems :sweat_smile:

PR #108, opened 2024-08-02
Include the GAMS version of the transport tutorial

glatterf42:
Essentially the same as #101, but with GAMS/gamsapi instead of linopy.

This PR is based on top of that branch, too, and actually adds only five commits. Two of those are the content of #107, so won't need to remain in here, either.

Hence: When reviewing/looking at the code, please focus on the new files in `tutorial/transport/`.

Please note: there is no test yet, but the values resulting from running the tutorial are correct. I'll add tests soon.


glatterf42:
This PR now includes exhaustive tests of the newly added code. The main thing I would like to improve for them is how paths are handled, but this will have to wait until we finalize a location for the gams/solve-related code.
The tests are currently failing because I didn't adapt the workflow file to install GAMS. This might also be influenced by where the solve-code ends up living; if it's not in this repo, we don't need to configure a secret holding a license here, etc. You can still convince yourself that the tests are running by checking out this branch locally on a system that has GAMS installed.

PR #109, opened 2024-08-06
adjust cli to upgrade databases from the manager

meksor:
title says it all. we need this for our first database migration that doesnt involve a persistent server

### DIFF
--- a/ixmp4/cli/platforms.py
+++ b/ixmp4/cli/platforms.py
@@ -176,7 +176,7 @@ def upgrade():
     )
     for m in manager_conf.list_platforms():
         if m.dsn.startswith("http"):
-            # This should probably never happen unless the manager registeres an
+            # This should probably never happen unless the manager registers an
             # external rest platform.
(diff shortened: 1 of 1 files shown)


ixmp4/cli/platforms.py
```diff
@@ -165,12 +167,24 @@ def list_():
...
+    utils.echo(
+        f"Establishing self-signed admin connection to '{settings.manager_url}'."
+    )
+
+    manager_conf = ManagerConfig(
+        str(settings.manager_url), SelfSignedAuth(settings.secret_hs256), remote=False
+    )
+    for m in manager_conf.list_platforms():
+        if m.dsn.startswith("http"):
+            # This should probably never happen unless the manager registeres an
```
danielhuppmann, line 179:
```suggestion
            # This should probably never happen unless the manager registers an
```

glatterf42 (approved):
LGTM :)

PR #114, opened 2024-09-25
set server log level to INFO

meksor:
We should probably make this configurable via an env var, but Im setting this down to INFO because we dont necessarily need debug info everywhere anymore.


glatterf42 (approved):
Okay, lgtm :)

PR #115, opened 2024-09-26
Configure `TestClient` to use internal error handling mechanism

glatterf42:
As noted by @meksor in [his review of #97 ](https://github.com/iiasa/ixmp4/pull/97#pullrequestreview-2327977670), raising `ValueError`s shouldn't work when working with the REST API. This PR disables this behaviour, so these PRs will need a little reworking and if we missed any such errors before, the checks on this PR will fail and we need some reworking here, too.

### DIFF
--- a/ixmp4/core/exceptions.py
+++ b/ixmp4/core/exceptions.py
@@ -178,12 +178,17 @@ class InvalidCredentials(IxmpError):
     http_error_name = "invalid_credentials"
 
 
+# == Optimization ==
+
+
+class OptimizationDataValidationError(IxmpError):
+    http_status_code = 431
+    http_error_name = "optimization_data_validation_error"
+
+
 # == Optimization.Table ==
 
 
-class OptimizationTableDataKeysNotUnique(NotUnique):
-    _message = (
-        "The keys of the Table's data are not unique. Please consider using "
-        "`constrained_to_indexsets` to specify the IndexSet identifiers."
-    )
-    http_error_name = "optimization_table_data_keys_not_unique"
+class OptimizationTableUsageError(IxmpError):
+    http_status_code = 432
+    http_error_name = "optimization_table_usage_error"
--- a/ixmp4/core/exceptions.py
+++ b/ixmp4/core/exceptions.py
@@ -182,7 +182,7 @@ class InvalidCredentials(IxmpError):
 
 
 class OptimizationDataValidationError(IxmpError):
-    http_status_code = 431
+    http_status_code = 422
     http_error_name = "optimization_data_validation_error"
 
 
@@ -190,5 +190,5 @@ class OptimizationDataValidationError(IxmpError):
 
 
 class OptimizationTableUsageError(IxmpError):
-    http_status_code = 432
+    http_status_code = 422
     http_error_name = "optimization_table_usage_error"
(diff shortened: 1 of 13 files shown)


meksor (approved):

ixmp4/data/db/optimization/base.py
```diff
@@ -16,6 +19,9 @@
 
 
 class BaseModel(RootBaseModel, mixins.HasCreationInfo):
+    # NOTE: only subclasses storing data actually define this!
```
glatterf42, line 22: Defining the type here so that I can access `BaseModel.DataInvalid` in the `utils.validate_data()` function. This might not be necessary though (we could also just raise `OptimizationDataValidationError`s straight from `ixmp4.core.exceptions` there) and it might be confusing since not all subclasses (like `Column`) actually define this error (since they don't all store and validate data). 
So please confirm this solution is fine or please suggest an alternative.

ixmp4/data/db/optimization/table/model.py
```diff
@@ -22,10 +24,12 @@ class Table(base.BaseModel):
     columns: types.Mapped[list["Column"]] = db.relationship()
     data: types.JsonDict = db.Column(db.JsonType, nullable=False, default={})
 
+    # TODO: should we pass self to validate_data to raise more specific errors?
+
     @validates("data")
     def validate_data(self, key, data: dict[str, Any]):
         return utils.validate_data(
-            key=key,
+            host=self,
```
glatterf42, line 32: Passing the whole instance at the moment to access both `.DataInvalid` and `__str__`, but if we change the way we raise errros in `utils.validate_data()`, we might also be able to just pass `__str__`.

ixmp4/data/db/optimization/utils.py
```diff
@@ -17,37 +19,43 @@ def collect_indexsets_to_check(
...
 
-def validate_data(key, data: dict[str, Any], columns: list["Column"]):
+def validate_data(host: base.BaseModel, data: dict[str, Any], columns: list["Column"]):
     data_frame: pd.DataFrame = pd.DataFrame.from_dict(data)
     # TODO for all of the following, we might want to create unique exceptions
     # Could me make both more specific by specifiying missing/extra columns?
     if len(data_frame.columns) < len(columns):
-        raise ValueError(
+        raise host.DataInvalid(
+            f"While handling {host.__str__}: \n"
```
glatterf42, line 28: Here's to hoping these lines will guide future users in identifying which objects cause the error.

ixmp4/core/exceptions.py
```diff
@@ -178,12 +178,17 @@ class InvalidCredentials(IxmpError):
     http_error_name = "invalid_credentials"
 
 
+# == Optimization ==
+
+
+class OptimizationDataValidationError(IxmpError):
+    http_status_code = 431
```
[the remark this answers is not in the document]
meksor: Hi, that's a good point. My favorite reference for this is https://http.cat/
This should probably be a (422)[https://http.cat/status/422]

ixmp4/data/db/optimization/utils.py
meksor: `host.__str__` should be a function...

ixmp4/core/exceptions.py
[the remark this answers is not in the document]
glatterf42: That's great, thanks for sharing :)

PR #116, opened 2024-10-01
Allow upgrading unmanaged local DBs

glatterf42:
As noticed by @meksor, it would be good to allow initializing locals DBs. Currently, this is only possible for managed DBs. This PR also allows upgrading/initializing unmanaged local DBs. Though please check: `settings.manager_url` and `settings.secret_hs256` have default versions, so we could also write:
```python
    utils.echo(
        f"Establishing self-signed admin connection to '{settings.manager_url}'."
    )
    manager_conf = ManagerConfig(
        str(settings.manager_url),
        SelfSignedAuth(settings.secret_hs256),
        remote=False,
    )
    platform_list = manager_conf.list_platforms() if settings.managed else settings.toml.list_platforms()
```
which might safe us from writing a test to cover these lines specifically, but introduces slight overhead costs for running `ManagerConfig().__init__` even for unmanaged instances. Not sure which is preferable here.

### DIFF
--- a/ixmp4/cli/platforms.py
+++ b/ixmp4/cli/platforms.py
@@ -167,24 +167,30 @@ def list_():
     )
 )
 def upgrade():
-    utils.echo(
-        f"Establishing self-signed admin connection to '{settings.manager_url}'."
-    )
+    if settings.managed:
+        utils.echo(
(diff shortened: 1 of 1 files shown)


meksor (approved):
I think this is good as it is, thanks!

PR #121, opened 2024-10-16
try to use NullPool to fix idle connection issue

meksor:
Currently, we keep running out of available connections on the postgres cluster because of how the scaling and connection pooling works. `NullPool` will probably slow us down a bit for each request but fix the issue for now until I can find a way to make this uncompromisigly scalable: https://docs.sqlalchemy.org/en/20/core/pooling.html#sqlalchemy.pool.NullPool


glatterf42 (approved):
Sorry for hijacking and stalling this PR. 

Just for posterity: the Code Quality check was broken because `ubuntu-latest` (the GitHub runner image) switched to 24.04, which comes with issues such as an externally managed Python, which does not allow you to `pip install` things unless you're in a venv. 
The pre-commit GitHub Action is in maintenance mode and unlikely to respond to that, so I fiddled around with running pre-commit [the recommended way](https://pre-commit.ci/). This needs to be enabled for the IIASA organization.
Pre-commit install a separate venv for every tool and is opinionated enough to never change that, regardless of how many people complain. Thus, for type checkers like mypy, workarounds need to be found to make them run in our poetry-managed venv. For this, https://jaredkhan.com/blog/mypy-pre-commit#option-1-use-language-system-to-run-mypy-in-an-existing-environment and https://github.com/python-poetry/poetry/issues/7091 were very useful, but try as I may, I couldn't get it to work.

In the meantime, enough push back has happened for Github to revert `ubuntu-latest` to 22.04, which works fine with the pre-commit action. Nevertheless, we should keep an eye open and try to find a more sustainable way of running mypy in our CI pipeline.

Issue #119, opened 2024-10-03
Make `optimization<item>.data` attributes consistent


glatterf42:
This mainly refers to `indexset.elements`, which should be called `indexset.data` for consistency. It is not actually important that the data take the form of "elements", which is itself a kind of arbitrary definition. But please also check if the other data attributes make sense.

glatterf42:
Check `core.opt.indexset.tabulate()` run-specificity, too.

PR #122, opened 2024-10-17
Normalize `IndexSet.data` DB storage

glatterf42:
Thinking about our DB design, I have come across [this description of data normalization](https://www.timescale.com/learn/guide-to-postgresql-database-design). This does not apply to the way we currently store data from optimization items, but it probably should. So this PR starts by normalizing the `IndexSet` storage, which is likely the easiest one.

First off, this PR renames `IndexSet.elements` to `IndexSet.data` to be consistent with the other optimization items, closing #119.

@danielhuppmann confirmed that the `.data` of any `IndexSet` will always be of the same `type()`, that's why we can move the `data_type` column to the `IndexSet` table and have a separate table for `IndexSetData`. This also allows us to drop a few test cases.

`data` is now stored always as a `string` because conversion from `string` to the other data types works best this way. This conversion happens in the [hybrid](https://docs.sqlalchemy.org/en/20/orm/mapped_attributes.html#mapper-hybrids) [property](https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html#join-dependent-relationship-hybrid) `.data`, which is not an official column, unfortunately, requiring the adaptation of the `tabulate()` method in the `IndexSetRepository`. 
If you immediately know how this or the type conversion can be made more efficient, please let me know :)

Other than that, all tests ~should~are still ~be~ passing, nothing in the user-facing API changed, and I haven't yet run any kind of stress test on this setup. Also, DB migrations are still missing.


ixmp4/data/db/optimization/indexset/repository.py
```diff
@@ -61,21 +61,36 @@ def list(self, *args, **kwargs) -> list[IndexSet]:
...
     @guard("view")
     def tabulate(self, *args, **kwargs) -> pd.DataFrame:
-        return super().tabulate(*args, **kwargs)
+        result = super().tabulate(*args, **kwargs).drop(labels="data_type", axis=1)
+        result.insert(
+            loc=0,
+            column="data",
+            value=[self.get_by_id(id=indexset_id).data for indexset_id in result.id],
+        )
+        return result
```
meksor, line 71: This looks quite slow

ixmp4/data/db/optimization/indexset/model.py
```diff
@@ -10,26 +8,50 @@
...
-            if element in unique:
-                raise self.DataInvalid(
-                    f"{element} already defined for IndexSet {self.name}!"
-                )
-            else:
-                unique.add(element)
-        return value
+    @db.hybrid_property
+    def data(self) -> list[float | int | str]:
+        return [cast_data_as_type(data, self.data_type) for data in self._data]
```
meksor, line 39: This also looks quite slow, but is porbably unavoidable?

ixmp4/data/db/optimization/variable/repository.py
```diff
@@ -65,7 +65,7 @@ def _add_column(
         self.columns.create(
             name=column_name,
             constrained_to_indexset=indexset.id,
-            dtype=pd.Series(indexset.elements).dtype.name,
```
meksor, line 68: ALso this looks quite expensive for what it does, no idea how to avoid it right now though...

glatterf42: For all items like `parameter`, `variable`, etc, I'm thinking that tables similar to `IndexSetData` might relieve us of the whole `Column` table. Especially if we end up translating `parameter.data` etc to `IndexSetData.id`s, since the main purpose of the columns is currently to store which indexset the data columns belong to and which type it has. All of this can probably be taken care of in a better way, eliminating these function calls :)
However, this will only happen in a later PR, I'm afraid.

ixmp4/data/db/optimization/indexset/model.py
glatterf42: Maybe @danielhuppmann can clarify, but I thought we want the `indexset.data` attribute to accurately portray the data type. So I guess we can either cast types here or in the API *and* core layers. 
But the function itself I was also wondering about: there should be some efficient built-in that does this kind of casting, my intuition tells me. Do you know of a better way? Would it be faster to collect all `self._data.value`s in a numpy array or so and cast that appropriately?

ixmp4/data/db/optimization/indexset/repository.py
glatterf42: Agreed, but how else would I do this? Unfortunately, sqlalchemy doesn't consider these hybrid properties to be data columns, so our `tabulate()` function, which just selects the whole tables, only returns the new `data_type` column and not the `data` attribute. Though also not the `_data` relationship, I think.
I'll double check if this can be configured in a more automatic way.

glatterf42: Would it be faster to adapt the `base` `tabulate()` function so it selects all indexsets while joining with `IndexSetData`?

ixmp4/data/db/optimization/indexset/model.py
glatterf42: According [to this guide](https://www.geeksforgeeks.org/python-type-casting-whole-list-and-matrix/), the `O()` of using loops and using numpy is both `O(n)`, but locally, the `tests/data` test runs 0.1 seconds faster when using numpy as it is used now, so hope that's better :)

ixmp4/data/db/optimization/indexset/repository.py
meksor: My first try would be to attempt to build sqlalchemy models form the data returned by tabulate directly or something like that.

ixmp4/data/db/optimization/indexset/model.py
meksor: Well of course they are both O(n) yes, doesnt really say much about the actual cost though.

meksor:
Hm so im unsure whether to try to get this PR in better shape or not @danielhuppmann. 
I feel like we are introducing a lot of technical debt right now, but i would need a month or two to weed out some problems, get to know the code fridolin added in the last year and so on. May be best to merge it for now and revisit when integrating toolkit and reducing the codebase with some big abstractions. I know this is basically going to come down to me cleaning up, which I am not a fan of, but Im kind of at a loss at what else to do.

ixmp4/data/db/optimization/indexset/model.py
glatterf42: I always thought it did, assuming the `n` can't be made smaller or so. Looking forward to your workshop on this :)

ixmp4/data/db/optimization/indexset/model.py
```diff
@@ -16,20 +16,34 @@ class IndexSet(base.BaseModel):
...
-                )
-            else:
-                unique.add(element)
-        return value
+    _data: types.Mapped[list["IndexSetData"]] = db.relationship(
+        back_populates="indexset"
+    )
+
+    @property
+    def data(self) -> list[float | int | str]:
```
glatterf42, line 26: I discussed how to do this best with an SQLAlchemy maintainer [here](https://github.com/sqlalchemy/sqlalchemy/discussions/12024). Since we are not going to use `.data` in SQL queries, we should be better served with a normal Python property.

ixmp4/data/db/optimization/indexset/repository.py
```diff
@@ -60,22 +60,45 @@ def list(self, *args, **kwargs) -> list[IndexSet]:
...
-            indexset.elements = indexset.elements + elements
+        if not isinstance(data, list):
+            data = [data]
+
+        bulk_insert_enabled_data: list[dict[str, str]] = [
+            {"value": str(d)} for d in data
+        ]
+        try:
+            self.session.execute(
+                db.insert(IndexSetData).values(indexset__id=indexset_id),
```
glatterf42, line 94: Using [ORM-enabled bulk inserting](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-bulk-insert-statements) should be faster than creating individual objects (as I did before).

glatterf42 (commented):
Some explanation for my recent changes. In addition, I read about partitioning, lambda caching, yield per, relationship loading styles, and some other topics which might further improve performance, but I don't want to keep this PR around too long to avoid scope creep.

ixmp4/data/db/optimization/indexset/repository.py
glatterf42: You recommended a flag to optionally load the `.data` from the relationship before, which I implemented now.

meksor: Right, that's the better approach sorry

glatterf42:
> I know this is basically going to come down to me cleaning up, which I am not a fan of, but Im kind of at a loss at what else to do.

I'm trying to avoid accumulating technical debt, so all that you see is there because I don't know better. I'm happy to clean up after myself, though, if you teach me how to write better code :)

meksor:
Alright the new changes make me feel a lot better about this PR, sorry for the smacktalk

meksor (approved):

glatterf42:
Thanks for the review! I understand that being in charge of multiple projects is consuming a lot of time and energy. I'm not offended by your comment, I only wish I could train my intuition to move towards where yours is faster. I'm happy to brainstorm ideas how we could do this :)

My last commits here target the DB migrations entirely. I forgot to include DB migrations with the PRs that introduced parameters, equations, and variables, and so initially two commits in https://github.com/iiasa/ixmp4/pull/101 added them back in. Since we'll likely merge this PR first and require a migration here, too, I've cherry-picked them to keep a linear migration history.

PR #124, opened 2024-10-18
Setup Codecov uploading and reporting

glatterf42:
When studying failed CI builds, I often find having to scroll past the codecov report tedious. In the rest of the MESSAGE stack, codecov is configured differently: a bot is picking up changes in coverage and reporting them as comments on the PR. For the rest of the stack, these are not just comments, but hard checks: should the PR reduce the test coverage (by too much), it will be rejected.
We don't have a policy like this with ixmp4 at the moment, so I'm not going to change that immediately. However, this PR tries to converge the different reporting styles: I have [enabled the codecov bot for this repo](https://github.com/organizations/iiasa/settings/installations/2244352), added a [`CODECOV_TOKEN` secret](https://github.com/iiasa/ixmp4/settings/secrets/actions), and configured codecov so that it should leave comments on PRs about coverage both within the changes proposed as well as the whole project in an informational way, i.e. not rejecting the PR because of lowered coverage.  

Please let me know what you think :)

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,14 @@
+coverage:
+  precision: 1
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
+
+comment:
+  require_changes: true
+  # Don't display the large graph
+  layout: "diff, files"
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -129,7 +129,7 @@ jobs:
       - name: Run tests
         run: |
           source .venv/bin/activate
-          pytest --backend ${{ matrix.backend }} --postgres-dsn "postgresql://postgres:postgres@localhost:5432/test" --cov-report xml:.coverage.xml --cov-report term --cov=ixmp4 -rsxX --benchmark-skip --color=yes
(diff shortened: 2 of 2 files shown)


glatterf42:
I think things are working as intended: the codecov checks for `patch` and `project` ran, but didn't produce a comment because the coverage is not changed by this PR.
Also, there's no base coverage to compare it to, yet, which would only be created by merging this PR. I infer this from the message [here](https://app.codecov.io/gh/iiasa/ixmp4/pull/124?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=iiasa), which already shows a coverage that corresponds to the latest terminal outputs I have in memory for this PR.
And the terminal output does not include the coverage anymore :)

In all likelihood, the comments produced on future PRs would not include as many details as the terminal output did. However, this could be [configured to our liking](https://docs.codecov.com/docs/pull-request-comments). Also, the dashboard on codecov (linked to from the individual PR checks) always includes more details per default.

meksor (approved):
Thanks, this looks great! I was waiting for the comment to appear but what you say makes sense and I'll look at it when the next pr commit comes In :)

PR #125, opened 2024-10-22
Upload coverage report from `main` directly

glatterf42:
While thinking about how to work with Codecov and the `squash & merge` workflow, I now believe the following workflow should work:

1. Get coverage report
2. Upload to Codecov as usual
3. Upload to Codecov again but with `override_branch: main` so that the commit also seems to belong to `main`

Subsequent PRs should then be able to find the reports (but see [here](https://github.com/codecov/feedback/issues/369#issuecomment-2428802464) for a caveat) and compare to them successfully.

So the first commit on this PR sets up the double uploading. Unfortunately, while thinking about this a bit more, I understood that this workflow won't work for PRs with revisions, so to speak: when the tests run once on a PR and upload its coverage, that coverage would be assumed to be valid for `main`, too. If you then need to improve your code later one and run the checks again, it would be run against the previously last commit of this PR, essentially rendering the `patch` checks (usually indicating coverage changes by this PR) meaningless.

Thus, the second commit simply enables the whole workflow on PRs as well as pushes to `main`, which wastes some computational resources, but seems to be the easiest way on our side.
I'll leave a note, however, to watch the issue linked above and revise our workflow once a better one becomes available.

On that note, why are we running the `build-docs` workflow on PRs and pushes to `main` (or at all, since we also let RTD build the docs for every PR)? Our own workflow is slightly faster than RTD, but the pages are not rendered as nicely.


meksor (approved):
Ok, I have no clue whats going on here exactly but it doesnt look like it will break anything... 

> On that note, why are we running the build-docs workflow on PRs and pushes to main (or at all, since we also let RTD build the docs for every PR)? Our own workflow is slightly faster than RTD, but the pages are not rendered as nicely.

Also no clue, the git blame points to me because I moved the workflow file at some point but i think originally this was added by @danielhuppmann  or @phackstock and might also be connected to the old self hosted docs setup?

PR #127, opened 2024-10-24
Fix Erroneous Audit Info Update

### LINKED ISSUE #126: Bug when setting the updated_at/by audit info
When editing a Run (adding/removing iamc-data or meta), the updated_at/by audit info entries are updated automatically. However, this update touches **all runs** with that model-scenario name, not just the run that was edited.

Steps to reproduce (I guess only relevant via Rest API):
- login as user A
- create a run
- login as user B
- create a run with the same model/scenario name
 
The first run (created by user A) now has "updated_by user B" in the audit info.

meksor:
As described by @danielhuppmann, when setting a run as the default, ixmp4 accidentally also updates that info for all other runs with the same model/scenario combination. 

This happens because the `set_as_default` function sets all the other runs to `is_default = False` before setting `True` on the singular subject run. The query is adjusted to only update rows which are actually not already `is_default = False` and I implemented a way to pause the event listeners to avoid setting that field on the previous default run as well. 

A repository can now pause the event listeners for a few lines like this: 

```py
with self.backend.event_handler.pause():
    # we dont want to trigger the
    # updated_at fields for this operation.
    self.session.execute(exc)
    self.session.commit()
```

Test is also adjusted to break if any of this ever happens again. 

fixes #126

### DIFF
--- a/ixmp4/data/db/events.py
+++ b/ixmp4/data/db/events.py
@@ -1,4 +1,5 @@
 import logging
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any, Mapping, Sequence, cast
 
 from sqlalchemy import Connection, event, sql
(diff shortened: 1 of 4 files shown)


danielhuppmann (approved):
This is very elegant! Thanks!

PR #129, opened 2024-11-12
Enable professional-grade mypy

glatterf42:
Finally, here it is!
After reading this blog post about "professional-grade" mypy, I began adopting the [professional-grade config](https://blog.wolt.com/engineering/2021/09/30/professional-grade-mypy-configuration/) for ixmp4. This PR is the result of the process.
It type hints everything, though some type hints might still need relaxation or tightening. Even our test suite is type-hinted here. This allows us to trust all variables we're dealing with have the correct type. In particular, refactoring can be done with improved confidence as mypy would catch a wide variety of bugs if the refactoring doesn't fit the existing code. Having these type hints also helps with readability, I think, as it is one kind of documentation in itself. I found this process especially insightful for understanding how the filters work :)
Of course, this PR also bumps mypy to its latest version.

That being said, there are a few points where I think we can still improve the PR, or add improvements through a future PR:

1. For the `create()` functions in the DB layer, I often kept the `*args` and `**kwargs` since the facade layer often uses positional arguments, while the rest/api layer uses keywords. This is different from `tabulate()` and `list()`, which are generally only called via `enumerate()` and with kwargs, so this would be a place to streamline the signatures.
2. In `data/abstract/meta`, I needed to type hint the `_type_map`, for which I needed to use the built-in `type`. This led to confusion for mypy since `meta` already had an attribute called `type`. I have renamed the attribute to `dtype` to enable the type hint.
3. For most `kwargs`, I have learned that the [best way to type hint them](https://mypy.readthedocs.io/en/stable/changelog.html#using-typeddict-for-kwargs-typing) is by creating a `TypedDict` subclass, which explicitly spells out all possible kwargs and their types. I have not done so for some `/base/` functions, which mainly act as templates or pass on all `*args, **kwargs` that they receive. When `args` or `kwargs` were unused, I sometimes removed them from the signature and type-hinted them as `Any` at other times, depending on how I understood the function's purpose. So there are currently two approached visible, which come down to different understandings of type hints, I suppose:
On the one hand, one could try to spell out everything explicitly how it is currently used. This improves the documentation aspect and enables mypy to catch more bugs before they reach production. On the other hand, this would block newly introduced kwargs until they are added to these kwargs-classes, which increases maintenance burden. Alternatively, one could convey the purpose of the function through the type hints, being more general. The pros and cons are essentially the inverse of the above.
For code that I didn't write, it's harder to correctly guess the purpose of functions, so I have generally opted more for the first approach. However, I'm open to discussion here.
4. Where I did write kwargs-classes, they are sometimes not organized as well as they could, I feel. In the DB layer, for example, `EnumerateKwargs` could often inherit from `CreateKwargs` at the moment. For the filter classes, spelling out everything like `name__like` is also lengthy, we could e.g. think of using mixins there.
5. For some `/base/` functions that pass on their arguments, I used `Any` because I couldn't get the alternative to work: in theory, this [sounds like a job](https://docs.python.org/3/library/typing.html#typing.ParamSpec) for `ParamSpec`: "They are used to forward the parameter types of one callable to another callable". However, I couldn't figure out exactly how to do that. On that note, if you know a good tutorial on `ParamSpec`, I'd be happy to read it as this almost seems like magic to me right now.

Lastly, I applied some `type: ignore` markers where I couldn't quite understand why mypy was complaining or where I intend to refactor the code very soon, anyway. In principle, though, almost none of that should be necessary, strictly-speaking. "Code I want to change soon" mainly refers to the DB model for parameters, equations, and variables, where I expect a more performant model can make do without the whole `column` thing. And I would maybe also drop `table` entirely since this is used nowhere in the tutorials. This is less certain, however, which is why I did type hint `table`, still.

>[!NOTE]
> As so nicely displayed by codecov, we don't have 100% test coverage yet (that's another juicy small project goal...). So when reviewing this, please use the branch like `main` might be used to ensure that things we don't test did not get broken. In particular, this probably means running some CLI commands from this branch.


glatterf42:
Another note: while working on this PR, I stumbled upon the setting `disallow_any_generics = true`. It is not active as of now, as that would introduce a few hundred more errors, but it would be an even stricter rule we could employ (to follow the spirit of using the strictest possible settings for mypy). This rule would prevent us from returning `dict` or `list` somewhere: all such generics would need to be explicitly specified as `list[int]` or some such. Happy to hear your thoughts on whether we should do that.

glatterf42:
If we adopt this PR, we should also present a [mypy badge](https://github.com/python/mypy/blob/master/README.md?plain=1#L11) in our readme :)

pyproject.toml
```diff
@@ -83,20 +83,30 @@ requires = ["poetry-core>=1.2.0", "poetry-dynamic-versioning"]
...
     '^doc\/',
     '^ixmp4\/db\/migrations\/',
 ]
 disable_error_code = ['override']
 implicit_reexport = true
-plugins = ['sqlalchemy.ext.mypy.plugin']
+plugins = ['numpy.typing.mypy_plugin', 'pandera.mypy', 'pydantic.mypy', 'sqlalchemy.ext.mypy.plugin']
+disallow_untyped_defs = true
+disallow_any_unimported = true
+# disallow_any_generics = true
```
glatterf42, line 94: Another note: while working on this PR, I stumbled upon the setting `disallow_any_generics = true`. It is not active as of now, as that would introduce a few hundred more errors, but it would be an even stricter rule we could employ (to follow the spirit of using the strictest possible settings for mypy). This rule would prevent us from returning `dict` or `list` somewhere: all such generics would need to be explicitly specified as `list[int]` or some such. Happy to hear your thoughts on whether we should do that.

meksor: If we are already at it and its a comparatively small amount of work I would tend to go ahead! "wenn schon denn schon"

ixmp4/data/abstract/iamc/timeseries.py
```diff
@@ -22,6 +29,21 @@ def __str__(self) -> str:
 ModelType = TypeVar("ModelType", bound=TimeSeries)
 
 
+class EnumerateKwargs(TypedDict, total=False):
+    id: int
+    id__in: Iterable[int]
+    region: dict[str, int | str | Iterable[int] | Iterable[str]]
+    unit: dict[str, int | str | Iterable[int] | Iterable[str]]
+    variable: dict[str, int | str | Iterable[int] | Iterable[str]]
+    run: dict[
```
meksor, line 38: Now that im seeing another one of these (this time for `run`): If we can somehow generate these `TypedDict` classes it should theoretically be possible to type this field with another one of those?

ixmp4/data/api/base.py
```diff
@@ -203,34 +215,38 @@ def _handle_response(
...
         json: dict | None = None,
-    ):
+    ) -> dict | list:
         """Convenience method for requests to the enumeration endpoint."""
         if params is None:
             params = {}
 
+        # TODO instead of assuming this, wouldn't it be better to never return None and
+        # handle this as an exception or some such in _request()?
+        # we can assume these types on enumeration endpoints
```
meksor, line 243: Hm, we return `None` if the response was successful but empty:
```py
            try:
                return res.json()
            except JSONDecodeError:
                if res.status_code < 300 and res.text == "":
                    return None
```

I feel like the repositories should be able to handle empty success responses, but im not sure if this is actually used somewhere right now...

meksor (commented):
Hi, what a feat! Thanks a lot, this is very cool.
I couldnt find what you mentioned in point 5. It sounds like that should be possible, but I've seen ParamSpec for the first time in this PR and couldnt tell you exactly what to do..
I also left some comments inline.

ixmp4/data/abstract/iamc/datapoint.py
```diff
@@ -35,6 +39,39 @@ def __str__(self) -> str:
         return f"<Datapoint {self.id} type={self.type}>"
 
 
+class EnumerateKwargs(TypedDict, total=False):
```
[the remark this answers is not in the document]
glatterf42: I'm not sure, I still have to read [mypy's docs on dynamic typing](https://mypy.readthedocs.io/en/stable/dynamic_typing.html). It seems to not support some dynamic features on purpose, being a static type checker and all.

ixmp4/data/api/meta.py
```diff
@@ -45,14 +48,27 @@ def get(self, run__id: int, key: str) -> RunMetaEntry:
...
-
-    def tabulate(self, join_run_index: bool | None = None, **kwargs) -> pd.DataFrame:
-        return super()._tabulate(json=kwargs, params={"join_run_index": join_run_index})
+    def list(
+        self,
+        join_run_index: bool | None = None,
+        **kwargs: Unpack[abstract.meta.EnumerateKwargs],
+    ) -> list[RunMetaEntry]:
+        # base functions require dict, but TypedDict just inherits from Mapping
+        json = {k: v for k, v in kwargs.items()}
```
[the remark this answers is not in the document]
glatterf42: I think I'm almost exclusively using `cast()` now, but will check again.

ixmp4/data/abstract/iamc/variable.py
```diff
@@ -89,7 +115,9 @@ def list(self, *, name: str | None = None, **kwargs) -> list[Variable]:
         """
         ...
 
-    def tabulate(self, *, name: str | None = None, **kwargs) -> pd.DataFrame:
+    def tabulate(
+        self, *, name: str | None = None, **kwargs: Unpack[EnumerateKwargs]
```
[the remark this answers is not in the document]
glatterf42: Done that and the same for all other `tabulate()` functions :)

glatterf42:
Okay, the update is here :)
It should clear the tests and mypy now, which has almost become `strict` (only missing `no_implicit_reexport`, which [mypy's docs](https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options) describe as "not too hard to get passing, but return on investment is lower".

I still need to do some cleanup, I want to look into your suggestion of recursive hints (and maybe employ some `TypeAlias`es) in particular, but right now, I'll have to take a step back and check if we can already support Python 3.13 :)
After I do the cleanup, I'll still raise any remaining questions explicitly.

ixmp4/data/api/base.py
glatterf42: Okay, I'm fine with keeping the `type: ignore` for now. I'll link this conversation for future reference, though :)

ixmp4/data/abstract/iamc/datapoint.py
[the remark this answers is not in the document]
meksor: Make sense, looks like there is a few (unresolved) discussions about this out there:
https://github.com/python/typing/issues/1495
https://github.com/pydantic/pydantic/discussions/7915
https://discuss.python.org/t/typing-unpack-for-kwargs-with-self-or-dataclass-classes/43001/8

I dont know if I really agree with the answer in the last thread. My goal is more to give developers IDE help than to introduce extra "safety" checks. 
But alas, seems this is not possible without some extra work, so screw it!

ixmp4/data/abstract/iamc/timeseries.py
meksor: See comment above, lets not go that far!

tests/data/test_count.py
```diff
@@ -1,11 +1,14 @@
...
 
 import pytest
 
 import ixmp4
+from ixmp4.data.backend import Backend
 
 
-def deepgetattr(obj, attr):
+# TODO How to type hint this to return Repos with .list() and .count()?
+def deepgetattr(obj: Backend, attr: str) -> Any:
```
glatterf42, line 11: This type hint is not great, but I don't know how to make it better.

tests/test_benchmark_filters.py
```diff
@@ -34,18 +37,23 @@
     ],
 )
 def test_filter_datapoints_benchmark(
-    platform: ixmp4.Platform, profiled, benchmark, filters
-):
+    platform: ixmp4.Platform,
+    # TODO: improve type hints for fixtures
+    profiled: Any,
+    benchmark: Any,
```
glatterf42, line 43: I haven't figured out how to properly annotate these fixtures. Do you know? If not, should we figure this out or accept them as they are now?

ixmp4/data/api/base.py
```diff
@@ -40,26 +60,63 @@ class BaseModel(PydanticBaseModel):
...
+                | float
+                | str
+                | dict[str, Any]
+                # TODO should be able to remove this once PR#122 is merged
+                | list[float | int | str]
+                | None
+            ]
+        ]
+        | None
+    )
```
glatterf42, line 119: This type hint is not great as explained in the comment. I think I really don't like pydantic using the order in type hints to coerce types, because that's something that I wouldn't expect. Here, it seems to be important that `int` come before `float`. Similar to `meta`, I tried using `StrictInt` etc, but this broke some other tests. Please let me know if you have a good idea for this.

glatterf42 (commented):
This took longer than expected again, but here it is: calling out every single thing I had a question about or an issue with! I'll push one or two more clean-up commits with things I noticed while going through all files, but after that, I'm happy with this PR :)

tests/test_benchmark_filters.py
meksor: profiled: probably need to make our own type alias in conftest.py since this is a custom fixture
benchmark: maybe `pytest_benchmark.fixture.BenchmarkFixture` works?

ixmp4/conf/manager.py
```diff
@@ -84,50 +95,46 @@ def _uncached_request(
         # NOTE: since this cache is not shared amongst processes, it's efficacy
         # declines with the scale of the whole infrastructure unless counteracted
         # with increased cache size / memory usage
-        res = self.client.request(method, path, *args, **kwargs)
+        res = self.client.request(method, path)
         if res.status_code != 200:
             raise ManagerApiError(f"[{str(res.status_code)}] {res.text}")
-        return res.json()
+        # TODO Can we really assume this type?
+        return cast(dict[str, Any], res.json())
```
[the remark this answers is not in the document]
meksor: I agree this is fine for now, especially since this might get replaced with scse-toolkit anyway.

ixmp4/core/exceptions.py
```diff
@@ -10,8 +13,17 @@ class ProgrammingError(Exception):
 ExcMeta: type = type(Exception)
 
 
-class RemoteExceptionMeta(ExcMeta):
-    def __new__(cls, name, bases, namespace, **kwargs):
+# TODO can't find a similar open issues, should I open one?
+# For reference, this occurred after adding disallow_subclassing_any
+# NOTE mypy seems to say Exception has type Any, don't see how we could change that
+class RemoteExceptionMeta(ExcMeta):  # type: ignore[misc]
```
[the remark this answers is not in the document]
meksor: OK, I suspect whats going on here is that mypy knows that the `Exception` class has no custom metaclass and thus subclassing from `type(Exception)` is the same as subclassing `type` directly.

As an explanation, in the python REPL we can see that: 
```
>>> type is type(Exception)
True
```
other classes with no custom metaclass yield the same result:
```
>>> from pandas import DataFrame
>>> type is type(DataFrame)
True
```
if they do have a custom metaclass however, like `DeclarativeBase`:
```
>>> from sqlalchemy.orm import DeclarativeBase
>>> type(DeclarativeBase) is type
False
```

So maybe removing the type alias `ExcMeta` and inheriting from `type` directly solves mypy's confusion?

ixmp4/core/iamc/variable.py
```diff
@@ -75,9 +72,8 @@ def tabulate(self, name: str | None = None) -> pd.DataFrame:
         return self.backend.iamc.variables.tabulate(name=name)
 
     def _get_variable_id(self, variable: str) -> int | None:
-        if variable is None:
-            return None
-        elif isinstance(variable, str):
+        # TODO this check seems kind of redundant...
+        if isinstance(variable, str):
```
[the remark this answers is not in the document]
meksor: Hm, so I think we are generally missing many argument checks for users without mypy. Since this is an internal function I would remove the check here completely and add checks in the user-facing core layer functions at some point(tm)

ixmp4/data/abstract/iamc/variable.py
```diff
@@ -23,6 +33,24 @@ def __str__(self) -> str:
...
+    # types, but won't actually know which strings to put in as the keys. Using the
+    # TypedDict, you'd know you have to look up its definition unless you know by heart
+    # what you can enter.
+    # run: dict[
+    #     str,
+    #     bool
+    #     | int
+    #     | Iterable[int]
+    #     | Mapping[str, int | str | Iterable[int] | Iterable[str]],
+    # ]
```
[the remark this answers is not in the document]
meksor: Agree.

ixmp4/data/db/filters/meta.py
```diff
@@ -19,5 +19,8 @@ class RunMetaEntryFilter(filters.BaseFilter, metaclass=filters.FilterMeta):
     value_float: filters.Float
     value_bool: filters.Boolean
 
-    def join(self, exc, **kwargs):
+    # TODO using specific form here as this seems to only be called on RunMetaEntry
+    def join(
+        self, exc: sql.Select[tuple[RunMetaEntry]], session: Session | None = None
+    ) -> sql.Select[tuple[RunMetaEntry]]:
```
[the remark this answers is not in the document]
meksor: I agree with crossing these bridges when we come to them.

ixmp4/data/db/meta/repository.py
```diff
@@ -151,68 +166,70 @@ def select_with_run_index(self) -> db.sql.Select:
...
         self.check_df_access(df)
-        df["type"] = df["value"].map(type).map(RunMetaEntry.Type.from_pytype)
+        df["dtype"] = df["value"].map(type).map(RunMetaEntry.Type.from_pytype)
 
-        type_: RunMetaEntry.Type
-        for type_, type_df in df.groupby("type"):
+        for type_, type_df in df.groupby("dtype"):
+            # TODO this should always be true, so better to avoid the check, but then
+            # mypy complains that we use a Scalar to access _column_map
+            assert isinstance(type_, str)
```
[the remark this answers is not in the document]
meksor: `cast`?

ixmp4/data/api/base.py
meksor: I would also guess that this worsens performance for the api client significanlty, did you run the benchmarks before and after? It will now check /every single value/ and try to coerce it, no?

tests/data/test_count.py
meksor: Well, use `object`, but that means you will always have to check the type of the return value afterwards.

ixmp4/data/db/optimization/variable/repository.py
```diff
@@ -84,6 +87,7 @@ def add(
 
     @guard("view")
     def get(self, run_id: int, name: str) -> Variable:
+        # TODO by avoiding super().select, don't we also miss out on filters and auth?
```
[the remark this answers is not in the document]
meksor: Yes, very true.

tests/data/test_count.py
glatterf42: We can do the same kind of type check with `Any`, which I have done now. This brought to light some missing kwargs for `count()`, which is great :)

ixmp4/data/api/base.py
glatterf42: There do seem to be differences, but I find it hard to put my finger on them. The `filter_` tests seem to run a little faster on `main`, but the other tests seem to be a lot faster on this branch.

tests/test_benchmark_filters.py
glatterf42: Benchmark won't be possible, unfortunately, until https://github.com/ionelmc/pytest-benchmark/issues/212 is fixed. I adapted the `TODO` and added a type alias for profiled as suggested :)

ixmp4/data/db/iamc/datapoint/repository.py
```diff
@@ -106,8 +124,12 @@ def join_auth(self, exc: db.sql.Select) -> db.sql.Select:
...
         return exc
 
-    def select_joined_parameters(self, join_runs=False):
-        bundle = []
+    def select_joined_parameters(
+        self, join_runs: bool = False
+    ) -> db.sql.Select[tuple[DataPoint]]:
+        # NOTE Not quite sure about this bundle, seems to possibly take all types of
+        # all model classes?
+        bundle: list[db.Label[str] | db.Label[int] | db.Bundle[Any]] = []
```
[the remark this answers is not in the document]
glatterf42: So `bundle` was a `list` already and I essentially just added the types that get added to it from the `extend()` calls below. `Model.name` would be a simple `str`, I think, but `Model.name.label("model")` returns a `Label[str]`, which is different for sqlalchemy: the `.label("model")` translates to `[SELECT ... FROM ...] AS model`. This is a different entity and thus requires the `Label` type.
I also don't think we can make a `Select[Bundle[...]]` work. Compare where this function is used: in `DataPointRepository.select()`, we craft an SQL statement as `exc`, which is either directly provided, coming from our function or created as `db.select(self.bundle)`. Since all these options are stored in the same variable, they should all have the same type. And `db.select(self.bundle)` simply returns a `Select[Tuple[Any]]` (where self.bundle is annotated generically as `Bundle[Any]`). Within `select_joined_parameters()`, we also don't return the `bundle`, but a `db.select(bundle)`, which turns into a `Select[Tuple[...]]`. 
Hope this makes sense and is fine :)

ixmp4/data/api/base.py
meksor: Hmm, so I tested my theory specifically and it seems to be true, not sure why it is not showing up in benchmarks here, maybe its a hint that its actually not such a big deal...

I took a few minutes to write this benchmark:

```py
import pytest
from pydantic import BaseModel

class Flat(BaseModel):
    data: list


class Detailed(BaseModel):
    data: list[int]


@pytest.mark.parametrize(
    "data",
    [
        list(range(1_000_000)),
        list(range(10_000_000)),
        list(range(100_000_000)),
    ],
)
def test_flat(benchmark, data):
    def bench():
        return Flat(data=data)
    
    benchmark(bench)



@pytest.mark.parametrize(
    "data",
    [
        list(range(1_000_000)),
        list(range(10_000_000)),
        list(range(100_000_000)),
    ],
)
def test_detailed(benchmark, data):
    def bench():
        return Detailed(data=data)
    
    benchmark(bench)
```
... which benchmarks a `list` type-hint versus `list[int]`.
The results are a little bit surprising:

```
------------------------------------------------------------------------------------------ benchmark: 6 tests -----------------------------------------------------------------------------------------
Name (time in ms)               Min                   Max                  Mean             StdDev                Median                IQR            Outliers       OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_flat[data0]             4.7867 (1.0)          6.0222 (1.0)          4.9039 (1.0)       0.1517 (1.17)         4.8734 (1.0)       0.0843 (1.0)          8;11  203.9214 (1.0)         149           1
test_detailed[data0]        17.1237 (3.58)        17.6601 (2.93)        17.3089 (3.53)      0.1298 (1.0)         17.2635 (3.54)      0.1706 (2.02)         19;2   57.7737 (0.28)         61           1
test_flat[data1]            65.7662 (13.74)       67.2023 (11.16)       66.1108 (13.48)     0.4609 (3.55)        65.9060 (13.52)     0.2877 (3.41)          3;3   15.1261 (0.07)         15           1
test_detailed[data1]       176.7581 (36.93)      180.5254 (29.98)      178.1051 (36.32)     1.4764 (11.38)      177.5801 (36.44)     2.2640 (26.85)         1;0    5.6147 (0.03)          6           1
test_flat[data2]           666.2161 (139.18)     674.3390 (111.98)     669.3321 (136.49)    3.3201 (25.58)      669.2435 (137.33)    4.9705 (58.95)         1;0    1.4940 (0.01)          5           1
test_detailed[data2]     1,797.9444 (375.62)   1,870.2783 (310.57)   1,817.4652 (370.62)   29.8766 (230.20)   1,808.0563 (371.00)   23.5364 (279.15)        1;1    0.5502 (0.00)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
**Both versions** scale with the length of the list, which I did not expect since in my head the "Flat" class should validate in constant time. (i now also wish i didnt call it "Flat")
Thats probably why it doesnt make a difference in light of the apparent performance improvements in this PR. (Do we know what these are?)

meksor (approved):

PR #131, opened 2024-11-20
Implement safeguard against problematic meta keys

danielhuppmann:
Just ran into an issue (with @pmussak) in the new ECEMF app and ixmp4 database instance - I had added meta-indicators named "version" to the database, which was "overwriting" the version-attribute in the run-selection component of the app.

To avoid such issues, this PR makes several terms illegal as meta-indicator-keys (at creation). I briefly thought about adding this to the meta-model (see link below), but I was worried that if a meta indicator with such a name was already created in a database with a current ixmp4 version, there might be issues with then loading them after upgrading ixmp4.

https://github.com/iiasa/ixmp4/blob/70366d4f1081d2a030ca8eb52feff2976a2b6724/ixmp4/data/db/meta/model.py#L64

### DIFF
--- a/tests/data/test_meta.py
+++ b/tests/data/test_meta.py
@@ -2,7 +2,7 @@ import pandas as pd
 import pytest
 
 import ixmp4
-from ixmp4.core.exceptions import SchemaError
+from ixmp4.core.exceptions import InvalidRunMeta, SchemaError
 from ixmp4.data.abstract.meta import RunMetaEntry
 
 from ..utils import assert_unordered_equality
@@ -36,6 +36,11 @@ class TestDataMeta:
             assert entry.key == key
             assert entry.value == value
             assert entry.type == type
+
+    def test_illegal_key(self, platform: ixmp4.Platform):
+        run = platform.runs.create("Model", "Scenario")
+        with pytest.raises(InvalidRunMeta):
+            platform.backend.meta.create(run.id, "version", "foo")
 
     def test_entry_unique(self, platform: ixmp4.Platform):
         run = platform.runs.create("Model", "Scenario")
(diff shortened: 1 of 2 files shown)


tests/data/test_meta.py
```diff
@@ -37,6 +37,11 @@ def test_create_get_entry(self, platform: ixmp4.Platform):
             assert entry.value == value
             assert entry.type == type
 
+    def test_illegal_key(self, platform: ixmp4.Platform):
+        run = platform.runs.create("Model", "Scenario")
+        with pytest.raises(InvalidRunMeta):
+            platform.backend.meta.create(run.id, "version", "foo")
+
```
meksor, line 44:
```suggestion
  def test_illegal_key(self, platform: ixmp4.Platform):
        run = platform.runs.create("Model", "Scenario")
        for illegal_key in {"model", "scenario", "id", "version", "is_default"}:
            with pytest.raises(InvalidRunMeta):
                platform.backend.meta.create(run.id, illegal_key, "foo")
```

Might be good to test all the keys from a distinct set here so we have to think about what we broke when removing keys from the actual set.

meksor (changes requested):
If I am not mistaken, a check in `RunMetaEntryRepository.bulk_upsert` is missing!

meksor:
Also not quite sure what exactly the problem is, i think the api currently does not allow for filtering my meta indicators... So is this check here now to account for something going on in the frontend implementation of the data explorer?

tests/data/test_meta.py
danielhuppmann: Added a loop for testing all illegal keys.

danielhuppmann:
Thanks @meksor, added the check also to `bulk_upsert()` (with a test).

For clarification, the aim of this PR is to prevent users from setting meta indicators that could be confusing, irrespective of whether it's confusing for the front-end API or for other users.

meksor (approved):
I see

PR #132, opened 2024-11-20
Support Python 3.13

glatterf42:
This PR adds support for Python 3.13, once that is possible. When I tried to do the same for pyam-iamc, poetry didn't let me do it because ixmp4 doesn't support it yet. Here, poetry had no complaints, so let's see :)

At the moment, I have just copied the tests for Python 3.12 and run them for 3.13 as well. This is not a very sustainable solution, though, and we should think about how long we want to keep supporting old Python versions.


glatterf42:
Both in the CI logs as well as locally, all these version bumps don't seem to accrue any new `DeprecationWarning`s. Thus, this PR seems good from my side :)

.github/workflows/pytest.yaml
```diff
@@ -27,6 +28,18 @@ jobs:
...
+            postgres-version: "16"
+            pandas-version: false
+            backend: "sqlite,rest-sqlite"
+          # pgsql 15
+          - python-version: "3.13"
+            with-pyarrow: false
+            postgres-version: "15"
+            pandas-version: false
+            backend: "postgres,rest-postgres"
+          # with pyarrow
```
meksor, line 42: I would say you can remove the python 3.12 postgres tests (Lines 49-53), if there is a specific issue with that combination we can turn them back on.

meksor (approved):
Looks good, one comment inline, but fine either way!

PR #140, opened 2024-11-28
Introduce `run.get_by_id()`

glatterf42:
The second part of cleaning up #101. This exposes the `run.get_by_id()` function to all remaining data layers as it is only used by `run.clone()` (see next clean up PR) in the data/server layers (and present already in the DB layer).


meksor (approved):
lgtm

meksor (approved):

PR #141, opened 2024-11-29
Introduce `run.clone()`

### LINKED ISSUE #133: Enable `Run.clone()`
In his review of #101, @danielhuppmann noted that it would be very user friendly to be able to call `Run.clone()` directly instead of having to call `runs.clone(run_id=...)`.

glatterf42:
Closes #133 :)
This is the next part of cleaning up #101, but it also improves the code that was originally introduced there, though I'm not entirely happy with all changes.

The goal was to enable `run.clone()` in the core layer. For that, `clone()` had to move outside of the core `RunRepository` since `Run` is not related to that. The most logical place to put it seemed `backend.runs.clone()` to me, so I did that, but it also warrants adding `clone()` to the REST API. 
With `clone()`, we're creating a new `run`, which sounds like a `POST` request to me. However, we already intercept all those requests to `/runs/`, so I had to find a new way. Rather than going for another request method, I opted to redirect all `POST` requests to `/runs/clone/`, which is not otherwise in use, to `clone()` in the DB layer. Please let me know if this is appropriate.
The actual logic for `clone()` worked well in the core layer, where it could make use of the `run.iamc` object encapsulating some validation and formatting operations. I tried to move these helper functions to the DB layer, but some require access to a specific `run`, making them unsuitable to be used in the DB `RunRepository`. Thus, I duplicated their logic within the `clone()` function, which is likely not the best solution. Please let me know which alternative we can employ.
Lastly, I found two oddities referring to type hints, both marked as `TODO` inline: 

- `core.indexset.data` returns a `list[float] | list[int] | list[str]` now, which is technically correct: there can only be one type in any `indexset.data` list. However, type checkers then don't know if they should allow something like `indexset.add("foo")` since they can't guarantee that `.data` is a `list[str]` this time. I'm not sure how, but it would be great to annotate `.data` ... dynamically I guess, so that it's recognized correctly and specifically.
- Some `iamc` functions like `iamc.datapoints.bulk_upsert()` are annotated inconsistently. In the abstract layer, they may proclaim to accept a `pd.DataFrame`, whereas in the DB layer, they require a `pandera.DataFrame` of a specific schema. This leads to some `type: ignore` that we could probably remove, but I'm not sure how we would want to do that.
  We could of course adapt the type hint to accept both kinds of dataframes, but we probably want to ensure that the data are validated even when the DB layer is called directly, so we probably don't want to let simple `pd.DataFrames` slip. Just using `pandera.DataFrame`s might be tricky, too, as e.g. `pandera.DataFrame["run__id"]` tells me I can't index a `pandera.DataFrame`. I'm happy to hear suggestions here :)


meksor:
I would not add an endpoint for this, especially as it does not actually add any functionality (by definition a run that is cloned could already be created via the REST API) and I hope that we wont be cloning thousands of runs on a regular basis.
The tests are very hard to read again, can we do something to make them easier to parse?

meksor:
Just to clarify: Ill accept the PR also with the current idea, just still reading through it...

danielhuppmann:
Two quick responses to @meksor:

> I would not add an endpoint for this, especially as it does not actually add any functionality (by definition a run that is cloned could already be created via the REST API) 

The benefit of a dedicated endpoint would be that the data would be copied within/close to the database, right? Avoiding that the data is sent back and forth via the RestAPI.

Also, having a dedicated `clone()` method avoids forgetting some parts of the Run (iamc-data, optimization-items, other stuff to be added in the future)

> I hope that we wont be cloning thousands of runs on a regular basis.

Sorry, cloning runs happens all the time in MESSAGE modelling - every time a user does a variation of an existing scenario, you start by cloning the base scenario and make some modifications.

meksor:
OK, convinced!

meksor (changes requested):
Hi, as said, i think the two tests added here could benefit from comments/whitespace/abstraction to make them easier to read. 
A hard requirement is a test that checks if `clone` respects the permission check! 
(It looks like it does since it uses `create`, but its best to bake this into a test.)
You can add a test in `test_auth.py`

glatterf42:
Thanks for your review :)
I'm not familiar with all the intricacies of the auth system, so please let me know if the test I added is already sufficient. Without the `if platform_info == self.gated:` clause, `run.clone()` fails on the `public` and `private` platforms due to missing permissions.

As for the readability: I tried adding some blank lines and comments, but I'm actually just adding one test here (not counting the auth test, which does not require the same setup), so I'm not sure how to abstract things away here short of a refactoring regarding how we call the methods of adding data to optimization items or listing them. Please let me know what you think.

meksor (approved):
Lgtm, and: I think this should continue to work after the versioning change as is.

PR #142, opened 2024-12-06
Increase apparent test coverage by removing superfluous ellipses

glatterf42:
[Yesterday I learned](https://stackoverflow.com/a/55275342/23037943) that one does not need to put any code inside a function that has a docstring. This allows us to remove numerous lines in the abstract layer that just contain `...` and are counted by test coverage as missing lines. Removing them should not affect any kind of behaviour or how the docs look, but increase the apparent test coverage :)


meksor:
Hmm, so what you are saying does check out. The ellipses is not really strictly required.

```py
from typing import Protocol

class A(Protocol):
    def f(self):
        """This is a docstring"""
        ...

class B(A): ...

B().f()
# ^ Cannot instantiate abstract class "B" with abstract attribute "f"
```
And alike:

```py
from typing import Protocol

class A(Protocol):
    def f(self):
        """This is a docstring"""

class B(A): ...

B().f()
# ^ Cannot instantiate abstract class "B" with abstract attribute "f"
```

**However** as soon as `A.f`contains a `yield` or `return` the mypy error *disappears*, so it might be worth leaving these few lines so it doesnt happen accidentally.

glatterf42:
I'm not sure I understand the problem with this. I created your example in the [mypy playground](https://mypy-play.net/?mypy=latest&python=3.12&flags=strict&gist=c0882ae753d6eba0a2d792df533f1f50), so please feel free to experiment with that.
The PR only removes `...` from functions where this is the only "body" of the function. Whenever a function contains `yield` or `return`, it remains unchanged. There is no intersection of `yield`/`return` and `...` in functions, it's either or. And the error is function-specific, as my playground example shows: `B().f()` doesn't work even if `f()` has a `return` statement if `g()` is abstract. So with or without the ellipses, one could only call the `yield`/`return` functions on child classes that actually provide implementations for the abstract functions.

Please let me know what I'm missing about this.

meksor:
Im not sure i understand the /benefit/ of this. IMO the ellipsis adds semantic value as in "this space intentionally left blank". 
Why are we removing them? To increase "apparent" test coverage, thats not a priority in my book. So I would not do it.

glatterf42:
Ideally, as I understand it, we should push for 100% test coverage (and more, since even calling each single line in the tests might not guarantee that it works as intended for various inputs). Codecov also counts lines containing only `...` as lines that require testing, so we have a few options of dealing with them in pursuit of the 100% coverage mark:

- We could tell Codecov that all such lines are to be excluded from measuring coverage (similar to `if TYPE_CHECKING:` lines). This might affect more lines than then ones I removed here, but maybe deservedly so.
- We could come up with some kind of test that calls these lines, but I'm not sure we should. There is no intention to ever call them, I think, so this could take the form of raising `NotImplementedErrors` when one tries, which we should (and easily could) then test again.
- However, I also learned that these ellipses are not actually required and we could remove them without losing any functionality. I went for this option because it seemed like the cleanest one to me.

We don't have to merge this version, of course. I think an empty function body conveys intent as well as `...`, but I can also instead include ellipses from coverage measurements.

meksor:
Maybe this is a solution that doesn't make us both unhappy: https://github.com/nedbat/coveragepy/issues/1042#issuecomment-1568779826

meksor (approved):

PR #143, opened 2024-12-17
Normalize `optimization.Table` DB storage

glatterf42:
This PR is a follow-up to #122: moving to the next optimization item, it redesigns the data model to normalize it and hopefully also make it more performant. 

For `Table`s, this is a bit more tricky than for `Indexset`s. `Table`s have multiple columns, each of which needs to be linked to an `Indexset` but can have a name other than the `Indexset`'s name. Each `Table` also needs to store data in table form, where different `Table`s will have a different number of columns. 
Previously, we solved this by dumping the data into a JSON field, but that did not achieve the desired performance, nor did it follow DB model normalization rules. 

In preparation of this PR, I first thought that our ideal model would look like this:

- `Table` and `Indexset` should get a many-to-many relationship. Sqlalchemy's [association proxy](https://docs.sqlalchemy.org/en/20/orm/extensions/associationproxy.html#module-sqlalchemy.ext.associationproxy) extension is an elegant solution for this, as it also offers to save optional fields (like distinct dimension names) along the relationship.
- `Table.data` only consists of values that are stored inside `Indexset.data` (that the `Table` is linked to), so ideally, we save just references to these values in a new table `TableData` and have a one-to-many relationship between `Table` and that. A `relationship` between that and `IndexSetData` seems like the ideal object for this.

Unfortunately, only the first point worked out. Sqlalchemy can't handle a `relationship` as part of a `UniqueConstraint`, preventing the `INSERT` workflow that this whole refactoring is aimed at achieving. Instead, the `TableData` is now storing the actual `str` form of the values, increasing storage requirements (compared to the ideal solution). This might be a source of future improvement.
To tackle the problem of varying numbers of columns, `TableData` is now hardcoded to store up to 15 values per row, most of which will always be `NULL`. This requires some clean up before returning the actual DB object like dropping `NULL` columns and renaming columns (to the name from either their indexset or specific `column_name`), which is likely the messiest part of this PR. Please let me know how that could be improved.

Locally, the tests run just as long as they do on `main`, but this is not proper benchmarking, which would be the actual test for this PR: it should handle a million values faster than `main` (and hopefully old ixmp). I previously tested this for `Parameter`s (which utilize upserting rather than just inserting), though, so I might have to draft another PR very similar to this one for them and run the manual benchmark again. A proper benchmark setup for optimization items (to run automatically) would be great (depending on how long we need to reach ixmp's performance).


glatterf42:
Just summarizing the status here: I still think we would benefit from normalizing the data-storage-part of the optimization tables, possibly using the approach contained here.
However, especially after #219, this branch will not be mergeable without much refactoring, so if the approach seems useful, it might be best to start a new branch and cherry-pick (via git or manually) the changes to that one.
In order to fully understand if this is a good idea, we'd need the spare time to come up with performance tests, though. Ideally testing:

- `tables.add_data()`
- `tables.remove_data()`
- and both options in ixmp/on JDBC for comparison

Then we could compare the JSON and normalized DB models, and also provide numbers to the MESSAGE team whether ixmp4 improves performance in this regard (noting that `Parameter.remove_data()` on JDBC is a very common pain point).

PR #146, opened 2025-01-09
Add the transport tutorial

danielhuppmann:
This PR copies over the transport tutorial (with linopy) from #101 in a more streamlined manner. Most importantly, the term "tutorial" is now used for the entire thing consisting of three notebooks:

0. set up the database
1. implement the base version of the transport problem
2. do a scenario/variation


glatterf42:
Not sure why mypy is still failing here. I bumped the version locally, which removed the linopy/numpy pin. So this latest commit might also break the usability of linopy, but poetry allowed this.

glatterf42:
Thanks for your updates here, @danielhuppmann! I have reworked some of them, mainly in terms of formatting: we seem to have different preferences how to highlight things. I differentiate between names of tools and software, which I **write like this**, and vocabulary of our tool in particular, including function names and code object names, which I `write like this`. I think this helps newcomers to spot the difference and they will need to learn the vocabulary, anyway. I also apply the same highlighting always to indicate that `Run` still means `Run` and not "run". The only exception to this rule is when I think we're talking about "real world" units, for examples, as opposed to the `Unit` objects we have in ixmp4.

tutorial/transport/2_transport-tutorial_variant.ipynb
```diff
@@ -0,0 +1,503 @@
...
+   "source": [
+    "run.version"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can check that the solution from [Notebook 1](1_transport-tutorial.ipynb) was indeed saved by inspecting an `ixmp4.Variable`.  \n",
+    "The `data` of any optimization object is stored internally as a dictionary."
```
glatterf42, line 133: Just noting here that this might change soon if the "Normalize Parameter DB storage" PR gets finished and accepted.

tutorial/transport/2_transport-tutorial_variant.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "## Cleaning up after doing this tutorial\n",
+    "\n",
+    "You can use the **ixmp4** Command Line Interface (CLI) to remove the database related to this tutorial.\n",
+    "\n",
+    "Run the following in the command-line:\n",
+    "```\n",
+    "ixmp4 platforms remove transport-tutorial\n",
+    "```\n",
+    "\n",
+    "The prompt will ask whether you also want to delete the SQLite database file from your computer, answer `y` for a complete deletion."
```
glatterf42, line 477: Maybe by having this cell here, we can remove the almost identical ones in Notebook 0.

danielhuppmann: Well, my thinking for having it twice is to show users that do something stupid during setup how to quickly restart - and then show how clean up at the end of the tutorial.

glatterf42: Sure, we can also have it twice :)

PR #148, opened 2025-01-13
Add profiling & benchmarks for `Indexset` and `Parameter`

glatterf42:
In order to evaluate the benefit of #143, @meksor asked me to write some profiling and benchmark tests for `parameters.add(data)`. This PR does exactly that and a bit more, so there are a few things to keep in mind:

- I started with writing tests for `Indexset` because `Indexset` already had its DB model normalized (#122). This might be helpful because we can already compare with this PR alone how the `Indexset` model compares to converting data to `dict`s and storing them as JSON. This comparison is not accurate, though, so my next task will be to duplicate #143 for `Parameter`. (We are interested in `Parameter` instead of `Table` because that needs the `UPSERT` functionality, the benchmark of which triggered this whole procedure.) On top of that branch, I can include exactly the same tests we have here for a proper comparison.
- This PR contains the script that creates the test data. This should probably not be committed to `main` or at least requires more cleanup before being committed.
- For now, some tests are not using the `big` data (because I only ensured that the tests are running locally, for which I didn't want to wait so long). For proper benchmarks runs, we may want to adapt this. And add some warmup-runs and iterations.
- This PR also contains `tests/fixtures/optimization/big/parameterdata.csv`, which is too large for GitHub's liking. When I pushed the commit adding the file here, I received the following:
  ```bash
  remote: warning: See https://gh.io/lfs for more information.
  remote: warning: File tests/fixtures/optimization/big/parameterdata.csv is 98.55 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
  remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
  ```
  I'm not sure if it's preferrable to use `git-lfs` or create the test data dynamically.
- This morning, I came across https://github.com/gazorby/fish-git-emojis?tab=readme-ov-file. I like the idea of enabling a quick overview of commits by using emojis and potentially enabling tools to run on commit messages, though I admit that "Optimization Profiling" is likely not the best scope to use. What do you think?


glatterf42:
Superseded by #150.

PR #149, opened 2025-01-16
Avoid superfluous DB calls from facade

glatterf42:
While working on #148, @meksor and I noted that there were always two calls to `validate_data` when adding data to a `Parameter`. This is superfluous, of course, and the culprit is in the core layer: in `parameter.add()`, we `add_data()`, but then also set `parameter._model.data`, which triggers another `validate_data()` run because that always fires when the attribute is set. 

This PR fixes that by setting `parameter._model` instead of just `.data`, also for all other relevant optimization items. This significantly boosts their performance.

### DIFF
--- a/ixmp4/core/optimization/equation.py
+++ b/ixmp4/core/optimization/equation.py
@@ -44,16 +44,16 @@ class Equation(BaseModelFacade):
         self.backend.optimization.equations.add_data(
             equation_id=self._model.id, data=data
         )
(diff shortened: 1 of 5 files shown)


meksor (approved):

danielhuppmann (approved):
Nice! Thank you!

I now remember that I noticed this issue a while back in the tests but didn't manage to solve it at the time...

PR #150, opened 2025-01-16
Add profiling & benchmarks for `Indexset` and `Parameter` -- clean

glatterf42:
In #148, we're adding a very large data file: `big/parameterdata.csv` has almost 100MB. In order to avoid bloating the git history of this repo, this PR was created loosely following the procedure outlined [here](https://github.com/iiasa/message-ix-models/pull/86): I created a clean clone of ixmp4 and rewrote its history to remove the very large file from it entirely. Then I merged the resulting branch back into a branch freshly started off of `main`, and I'm using this to open this PR.

The preview shows 116 commits, but it's actually just one, merging the clean branch to the new one. The files changed seems reasonable, though, so I figure we should be fine by squashing these apparent commits when merging.


glatterf42:
Superseded by #155.

PR #155, opened 2025-01-30
Add profiling & benchmarks for `Indexset` and `Parameter` -- clean

glatterf42:
This PR adds profiling tests for `Parameter`s and `Indexset`s. It also increases performance by avoiding the `copy.deepcopy()` calls when validating optimization data and adjusts the validation function accordingly.

Unfortunately, #150 has fallen out of sync with `main`. When trying to do a `git rebase` there, git wants to rebase _all_ commits, which in this case means all 118 commits in the whole repository, creating lots of unnecessary conflicts. 
So instead of dealing with that, I looked at the largest files in our git history (thanks to [stackoverflow](https://stackoverflow.com/questions/9456550/how-can-i-find-the-n-largest-files-in-a-git-repository)) in a fresh clone and found that it doesn't contain the big parameterdata file that was committed by accident. 

<details>
<summary>List of ten largest files in our git history right now</summary>

```bash
d305e273e990  425KiB tests/fixtures/optimization/medium/parameterdata.csv
b781df98acee  716KiB tests/fixtures/optimization/medium/parameterdata.csv
a07bfd345778  996KiB tests/fixtures/medium/datapoints.csv
92fe972a2a27  2,2MiB tests/fixtures/big/parameterdata.csv
6716fc873ab2  6,5MiB tests/fixtures/big/datapoints.csv
38d78752a53b  6,6MiB tests/fixtures/optimization/big/indexsetdata.csv
6f4595a13944   15MiB tests/fixtures/optimization/big/indexsets.csv
e2a9ff495a0b   16MiB tests/fixtures/optimization/big/parameters.csv
d8a3bb70b4d4   32MiB tests/fixtures/big/upsert_parameterdata_vk.csv
762b69cd8c81   32MiB tests/fixtures/big/insert_parameterdata_vk.csv
```
</details>

Which means: the branch was deleted and never merged, thus the file was never present and we don't need to invoke `git filter-repo` to clean up. All we need to do is open a new PR based on `main` and cherry-pick the commits we're interested in, which is what this PR does.

> [!NOTE]
> The two largest files come from the branch #112, which will likely not be merged anytime soon. It also contains the script with which these files were created, so we can already cut repo size quite significantly by going to that branch and removing these files (first try a new commit doing that, then checking files sizes in a fresh clone again, if that didn't work, re-work the commit that introduced the files | alternatively, delete this branch, but I'd like to keep it for now in case we want to revisit that topic).


meksor (approved):
lgtm

PR #157, opened 2025-01-31
Remove `Column` class entirely

glatterf42:
As discovered in #151, using sqlalchemy's associationproxies is both (a bit) more performant and (considerably) less confusing than having a DB table called `Column`. Thus, this PR completes the removal of this construct.

TODO

- [x] DB migrations, though ideally, #156 gets its own migrations first

On top of this, the PR includes a few minor fixes:

- Make type hints of `.data` in the core layer more accurate
- Fix some `type: ignores` by casting appropriately
- Fix some docstrings wrt the errors they raise
- Make `Variable.indexsets` `None` when appropriate
- Fix missing `.set_creation_info()` calls
- Skip data validation iif appropriate (we didn't skip checks for e.g. variable.data with only "levels" and "marginals" after #139 and never skipped for tables, even when `{}` was added)


ixmp4/server/rest/optimization/variable.py
```diff
@@ -25,7 +23,7 @@ class VariableCreateInput(BaseModel):
 
 
 class DataInput(BaseModel):
-    data: dict[str, Any]
+    data: dict[str, list[float] | list[int] | list[str]]
```
glatterf42, line 26: I changed this to reflect the other changes, then run into the thing that the tests on the REST API platforms always converted data to `dict[str, list[float]]` for some reason. I don't understand why, but it doesn't seem to be this change because the same happens for equations. Still, to show that, I didn't also change the other type hints in this layer, though I suppose I still could.

ixmp4/data/db/optimization/equation/model.py
```diff
@@ -17,19 +32,40 @@ class Equation(base.BaseModel):
...
+        "_equation_indexset_associations", "column_name"
+    )
+
+    @property
+    def indexsets(self) -> list[str]:
+        return [indexset.name for indexset in self._indexsets]
+
+    @property
+    def column_names(self) -> list[str] | None:
+        return cast(list[str], self._column_names) if any(self._column_names) else None
```
meksor, line 69: What is the rationale behind the one list being `None` if empty?

meksor: `list(self._column_names.all_())` could work for the list conversion as well?

meksor:
Looks alright to me, does it depend on https://github.com/iiasa/ixmp4/pull/156 ?

ixmp4/data/db/optimization/equation/model.py
glatterf42: I don't think `._column_names` has a `.all_()` attribute. There is an `all_()` function in sqlalchemy, but that seems to produce an SQL `ALL` expression, so I'm not sure how that's useful here.

glatterf42:
It doesn't depend on #156, but it's just as well that that was merged first :)

ixmp4/data/db/optimization/equation/model.py
meksor: And I have to reiterate my concern about the possibility of this being None, it seems spurious to me that one has to constantly check `if X is None` so the type checker doesn't complain instead of always treating it like list...

meksor: I think thats because sqlalchemy doesnt understand that this will become an `AssociationProxyInstance` instead of and instance of `AssociationProxy`, but i might be wrong

glatterf42: And I think the case where there are no `column_names` other than the `indexset_names` is at least as important as the case where there are some. So I'm not sure how much benefit storing an empty list would bring. 
Mainly, I'm using this attribute to make sure the correct data are validated, for which I need to know the complete list of column names, which I can easily do via `names = column_names or indexset_names`, which is not that different from `names = column_names if len(column_names) else indexset_names` or even shorter. So I'm not sure that `[]` is better in the practical cases we already have.

meksor: With a list it works the exact same way...
```py
>>> [] or ["foo"]
['foo']
```

meksor: And I'm not sure breaking convention with how any other list works in the whole programming language is a good way to save ~10 characters in some internal function??

ixmp4/data/db/optimization/table/model.py
```diff
@@ -1,14 +1,24 @@
...
+
+
+class TableIndexsetAssociation(base.RootBaseModel):
+    table_prefix = "optimization_"
+
+    table_id: types.TableId
+    table: types.Mapped["Table"] = db.relationship(
+        back_populates="_table_indexset_associations"
+    )
+    indexset_id: types.IndexSetId
```
meksor, line 18: ditto

meksor (changes requested):

ixmp4/data/db/optimization/equation/model.py
meksor: >  In any case it doesn't have an append function, nor will it work with len() (educated guess) so essentially we are misleading anyone who tries to use these functions on this so called "list" ;)
Sorry, just for clarification, this is wrong. I think I am getting mixed up between the functionalities of the django version of this type of mapping and misreading the sqla docs.

However, being confused by the new version I have to admit that my feedback was uneducated, since if I now understand correctly: 
`column_names` is optionally provided when making a new equation/parameter/... object. If `column_names` is supplied it must be of the same length as `constrained_to_indexsets` and will act as a "key" for the columns instead of the "indexset name". From this perspective returning the value supplied to the creation function kind of makes sense, however pulling this info back out of the association table like this was confusing me and I wasnt able to make the connection.
I am wondering if it is necessary to be able to distinguish if an equation has been supplied with overridden "column_names" later. Depending on if thats the case it might be beneficial to /always/ store something in `column_name` on the association table. For one, this eliminates the need for two properties, and it will also address another potential problem:
It is unclear to me how the association_proxy orders list items. By default postgres will return items in a practically random ordering, which /could mean/ that when cloning via: 
```py
self.backend.optimization.equations.create(
      run_id=run.id,
      name=equation.name,
      constrained_to_indexsets=equation.indexset_names,
      column_names=equation.column_names,
  )
```
...the indexsets and column names are suddenly switched! I can unfortunately not find out if this is a problem in practice since there does not seem to be a test case that clones a run with an equation like that. Trying to adjust `test_run_clone` to cover the case reveals that adding data to an equation with `column_names` throws an error and in turn that there is no test covering that either: 

```
ixmp4/data/db/optimization/equation/repository.py:148: in add_data
    data.set_index(index_list).combine_first(existing_data).reset_index()
TypeError: The parameter "keys" may be a column key, one-dimensional array, or a list containing only valid column keys and one-dimensional arrays.. Received column of type <class 'sqlalchemy.ext.associationproxy._AssociationList'>
```
Which brings me back to a complaint from earlier in this thread: This `cast()` call is not enough for any run-time type checks and pandas seems to not like that very much.

meksor: I also just noticed that there is a relatively or PR already addressing this problem! #162

glatterf42: Okay, I've opened a new issue about the test case you reported, and I'll push a commit that adds `order_by` to the `*_indexset_associations` relationship. Please check to see if that's all we need for now :)

Not sure why the tests are not running for https://github.com/iiasa/ixmp4/pull/157/commits/7512142753e92647053fde0c6fb5c386eb242399, but the checks for [818b782](https://github.com/iiasa/ixmp4/actions/runs/14857025963/job/41712694986) and [098f586](https://github.com/iiasa/ixmp4/actions/runs/14857062487/job/41712815553) are passing (and https://github.com/iiasa/ixmp4/pull/157/commits/7512142753e92647053fde0c6fb5c386eb242399 fixes the one code style test that was broken somehow).

PR #158, opened 2025-02-03
Introduce `Indexset.remove_data()`

glatterf42:
For iiasa/ixmp#557, I will next need functions that allow removal of individual keys from Indexsets, Tables, and Parameters. This PR adds such a function for Indexsets.

While working on that function, I found myself thinking: wouldn't it be nice to get some sort of warning if you're trying to remove data from an Indexset, but that Indexset doesn't have that data? This can be checked easily in sqlalchemy, so I added `log.info()` messages for two cases. 
In order to test that these messages are actually logged, I copied over the `assert_logs()` utility function we have in ixmp and enhanced it with some type hints. Thanks to @khaeru for writing that in the first place!

> [!NOTE]
> This PR looks larger than it is because I swapped the order of the tests in the `data/` subfolder to be consistent with the other optimization tests. There's no need to verify `test_list_indexset` and `test_tabulate_indexset` again.

### DIFF
--- a/ixmp4/server/rest/optimization/indexset.py
+++ b/ixmp4/server/rest/optimization/indexset.py
@@ -54,12 +54,24 @@ def create(
 
 
 @autodoc
-@router.patch("/{indexset_id}/")
+@router.patch("/{indexset_id}/data/")
 def add_data(
     indexset_id: int,
     data: DataInput,
     backend: Backend = Depends(deps.get_backend),
 ) -> None:
     backend.optimization.indexsets.add_data(
+        indexset_id=indexset_id, **data.model_dump()
+    )
+
+
+@autodoc
+@router.delete("/{indexset_id}/data/")
+def remove_data(
+    indexset_id: int,
+    data: DataInput,
+    backend: Backend = Depends(deps.get_backend),
+) -> None:
+    backend.optimization.indexsets.remove_data(
         indexset_id=indexset_id, **data.model_dump()
     )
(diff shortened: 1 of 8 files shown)


ixmp4/server/rest/optimization/indexset.py
```diff
@@ -63,3 +63,15 @@ def add_data(
...
     )
+
+
+@autodoc
+@router.delete("/{indexset_id}/")
+def remove_data(
+    indexset_id: int,
+    data: DataInput,
+    backend: Backend = Depends(deps.get_backend),
+) -> None:
```
meksor, line 74: Can we move this and `add_data` (`PATCH /{id}/`) to `DELETE /{id}/data/` and `PATCH /{id}/data/` respectively? Not sure how the other endpoints are laid out and if they need to be adjusted as well: Currently `DELETE /{id}/` is used in lots of places to remove the object entirely so if this functionality is needed for `IndexSet`s as well, we would have to make the endpoints `DELETE /{id}/self/` or something like that which is inconsistent and inelegant.

meksor (changes requested):

ixmp4/server/rest/optimization/indexset.py
glatterf42: Done, though for Tables and Parameters, this change will be included in #159 and #160, respectively.

meksor (approved):
ill approve with a note to consider making the tests similarly easy to understand as per the response in another PR, although here i get the gist i think.

PR #159, opened 2025-02-04
Introduce `Table.remove_data()`

glatterf42:
As announced in #158, this PR adds `Table.remove_data()` (and fixes some `type: ignore` comments).

Two things to note here, mainly:

- It would be ideal to find a way to skip the data validation when removing data. As the code is written now, `table.data = ...` triggers the data validation, but when we're removing data, we don't actually need to validate the data remaining in the `table` as that has been validated when we added it previously. 
  Unfortunately, sqlalchemy's docs don't hint at any possibility to skip the validation check when using `@validates` as we do. They do state that removals from a collection generally don't trigger this, but we don't have a proper collection, just a `dict` data attribute. So this would be a benefit of normalizing our data model further, since `table.data` would then be a proper collection.
  In the meantime, though, I was wondering whether we need `@validates` at all. We do need to validate data, but we don't expose `table.data` in the DB layer, so there's no chance of a user accidentally setting this to some invalid data. Thus, I'm not sure we benefit all that much from the `@validates` functionality of automatically checking `.data` each time it is set; we could simply make sure that `add_data()` always calls our validation function manually -- and conversely, that `remove_data()` never does that.
  What do you think?
- I'm not entirely sure why `remove_data()` doesn't seem to need a `self.session.commit()`, while `add_data()` does need it to pass the tests. I'm not sure we need to worry about this, either, though, as I'm happy as long as the test work as intended.


ixmp4/server/rest/optimization/table.py
```diff
@@ -67,6 +67,16 @@ def add_data(
     backend.optimization.tables.add_data(table_id=table_id, **data.model_dump())
 
 
+@autodoc
+@router.delete("/{table_id}/data/")
```
meksor, line 71: Same here
```suggestion
@router.delete("/{id}/data/")
```

tests/core/test_optimization_table.py
```diff
@@ -232,6 +232,55 @@ def test_table_add_data(self, platform: ixmp4.Platform) -> None:
         table_5.add(data={})
         assert table_5.data == test_data_5
 
+    def test_table_remove_data(self, platform: ixmp4.Platform) -> None:
```
meksor, line 235: Can we make this test easier to understand? Test data is being reused, some of the behavior is only underandable when one understand the behavior of the builtin `remove` function, there is lots of back and forth with `.set_index()`,`.index` and `.reset_index()`. I basically have to decipher this test line by line to get an idea of what the intended feature is actually doing.

meksor (changes requested):

tests/core/test_optimization_table.py
glatterf42: Hm, I just took a look at this and added some comments. But I guess this is not what you intended?
Basically, I wanted to avoid hardcoding the test expectations, so I tried creating them based on the original test data input. However, this may not be a very good test: if something in the pandas `.isin()` logic (or so) changes overnight (which I don't consider likely, but you never know), the test expectation will change just like the resulting data in the background, so we wouldn't catch that. So maybe I have to come up with a different way of creating the expectation data here, is that what you mean, too?

meksor: Kind of, yes. What is the argument against hard-coding test data? Having three hard-coded objects `intial_data` `remove_data` and `expected_data` seems like it would make a more robust and understandable test at the same time. 
I'm less worried about the behaviour changing and more just having a hard time reading the test fluently.

glatterf42: Only maintainability; if the input data need to change at some point, the hardcoded expectation also needs to change. But I agree: the way it is now is probably easier to read and should even catch behaviour changes :)

PR #160, opened 2025-02-04
Introduce `Parameter.remove_data()`

glatterf42:
Exactly the same as #159: this PR adds the `Parameter.remove_data()` functionality. Please see #159 for additional notes, as they apply to both PRs equally.


tests/core/test_optimization_parameter.py
```diff
@@ -239,6 +239,60 @@ def test_parameter_add_data(self, platform: ixmp4.Platform) -> None:
...
+            IndexSet(_backend=platform.backend, _model=model)
+            for model in create_indexsets_for_run(platform=platform, run_id=run.id)
+        )
+        indexset_1.add(data=["foo", "bar", ""])
+        indexset_2.add(data=[1, 2, 3])
+        test_data_1: dict[str, list[int] | list[str]] = {
+            indexset_1.name: ["foo", "foo", "foo", "bar", "bar", "bar"],
+            indexset_2.name: [1, 2, 3, 1, 2, 3],
+            "values": [1, 2, 3, 4, 5, 6],
+            "units": [unit.name] * 6,
```
meksor, line 255: What happens if I try to remove `{"values": [1]}` or `{"units": ["Unit"]}`?

meksor (changes requested):
All my feedback to #159 also applies here, one additional question though.

tests/core/test_optimization_parameter.py
glatterf42: Just to clarify: after the test setup, you are interested in what happens when running something like

```python
        data = {"units": ["Unit"]}
        parameter.remove(data=data)
```

At the moment, the `remove_data` functionality works via pandas. We take the existing data and set all indexed columns (i.e. those linked to an IndexSet, of which there is always at least one for Parameters) as the index. Then we do the same for the to-be-removed data. This is because what we actually want to remove are "value" and "unit", but we cannot just look for matching values because each "value" and "unit" might be present many times.
So by way of comparing the indexes, we then determine which data is supposed to remain. Technically, this is achieved by setting the remaining data to the existing data where the index of the existing data is not in the to-be-removed data.
So specifying just `{"values": [1]}` (for example) is not going to work because we could never know which `1` to delete. In practice, this currently results in a KeyError:

```python
>       parameter.remove(data=data)

tests/core/test_optimization_parameter.py:281: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ixmp4/core/optimization/parameter.py:50: in remove
    self.backend.optimization.parameters.remove_data(id=self._model.id, data=data)
ixmp4/data/auth/decorators.py:44: in guarded_func
    return func(self, *args, **kwargs)
ixmp4/data/db/optimization/parameter/repository.py:181: in remove_data
    data.set_index(index_list, inplace=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self =   units
0  Unit, keys = ['Indexset 1', 'Indexset 2']

...

E           KeyError: "None of ['Indexset 1', 'Indexset 2'] are in the columns"

.venv/lib/python3.12/site-packages/pandas/core/frame.py:6122: KeyError

```

This question led me to consider something else, though: we do not actually need the to-be-removed data to contain "values" and "units" columns/keys because removal is based solely on the indexed columns. So if you supply just `data = {indexset_1.name: ["foo"], indexset_2.name: [1]}` to `remove_data()`, the removal is successful. 
Also, when you remove that same data again, it is still successful in that it doesn't throw an error: pandas' `isin()` function will simply determine that none of the existing data's index values are in the to-be-removed data's index values and set the remaining data to the complete existing data.

What do you think should happen?

meksor: Hm okay, I hope I'm following correctly:
There is no reason to try that since the entire purpose of this function is to specify one or multiple rows of the "complete index" for which also "unit" and "value" data will be removed (since leaving them would make the columns of inconsistent length, beside not making any sense).
That brings up another question: what happens if you specify an 'incomplete' index? Eg. Just `{"Indexset 1": ["foo"]}`?
Also KeyError I assume?
We need to get rid of the KeyError because it will become a 500 Error Response since it does not inherit from the jsonable error class. While doing that you could provide a more informative error message "multiple candidates found for removal" and "specify one or more complete indices to remove associated units and values".

glatterf42: Hm, I suppose there are more ways to trigger errors that are not jsonable (though I don't have an example right away). Do we claim that any and all errors raised in the DB layer will make it through the API layer? It seems to me that this would require strict control of our inputs, which I'm not sure we currently do.
But for this case, we can of course catch and convert the error (and include a test to cover this). Do you want this done as part of this PR or could we postpone that to after we merge this and #171 (very soon, I hope)?

meksor: Well yes, unless they are caught of course. 
I would say yes, please adjust this PR at least for the things we found in this thread.

glatterf42: Done, also for `Table` and `IndexSet` :)

meksor (approved):

tests/core/test_optimization_parameter.py
glatterf42: Though generally speaking, the logic we use to remove data for `Table`s and `Parameter`s is pretty similar. In theory, one test of that logic should be enough.
Also, the tests of the core layer are (of course) in large parts running the same logic as the tests of the data layer, so maybe there's some way to reduce runtime there, too?

PR #162, opened 2025-02-05
Fix actual return type of `column_names` property

glatterf42:
Uncovered another bug introduced by #156 and #157: previously, I returned the `column_names` property as a `cast(list[str], ...)` of the underlying sqlalchemy collection. Mypy understood this just fine, but the actual type of the return object was not changed, it was just made to look that way. This created an error in `pd.DataFrame().set_index()` as this function does not know how to handle sqlalchemy's `_AssociationList` objects. 
This PR fixes this bug by explicitly creating the return object as a `list[str]`. The `for` loop should be fine as no item should have more than 12 `column_names` in practice. It also adds tests to ensure this behaviour doesn't break again in the future.


meksor (approved):
This makes me feel much safer than the version in #157 !

PR #163, opened 2025-02-11
Introduce optimization item `delete()`

glatterf42:
One more piece we need to complete functionality also offered by ixmp: this PR adds `delete()` functions for all optimization items as a whole.

I tried to achieve this using [sqlalchemy's cascades](https://docs.sqlalchemy.org/en/20/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships), which are recommended as [the higher-performing option in their docs](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#deleting-rows-from-the-many-to-many-table) (compared to loading all elements from collections before `delete()`ing them piecewise). 

> [!NOTE]
> While the docs say this requires [special setup for SQLite](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#sqlite-foreign-keys), we already have this exact setup, so can employ this immediately.

With the current setup, `IndexSet`s cannot be deleted if they are used by some other object. IIRC, this is a requirement that @danielhuppmann mentioned at some point, which we seem to get almost for free due to our setup: the `IndexSet.id` columns in the e.g. `TableIndexSetAssociation` cannot be `NULL`; thus, when one tries to delete an `IndexSet` while it is linked to a `Table` (i.e. recorded in `TableIndexSetAssociation`), an `IntegrityError/DeletionPrevented` will be raised.

All other optimization items should also theoretically be able to raise `DeletionPrevented`, but in practice, no other item relies on their `id`s not being `NULL` somewhere, so I'm not sure how else to check this. This is the only kind of `TODO` mark appearing inline. 
However, I'm also not sure we need to check this as the same kind of behavior is checked by the `IndexSet` tests or may not be relevant for the other objects.

### TODO

- [x] This PR changes some of the db/model files, but only relationship and foreignkey setup, so I'm not sure it needs its own migration files. But check if it does!


ixmp4/server/rest/optimization/scalar.py
```diff
@@ -75,3 +75,12 @@ def create(
     backend: Backend = Depends(deps.get_backend),
 ) -> Scalar:
     return backend.optimization.scalars.create(**scalar.model_dump())
+
+
+@autodoc
+@router.delete("/{scalar_id}/")
+def delete(
+    scalar_id: int,
```
meksor, line 83:
```suggestion
@router.delete("/{id}/")
def delete(
    id: int,
```

ixmp4/server/rest/optimization/table.py
```diff
@@ -84,3 +84,12 @@ def create(
     backend: Backend = Depends(deps.get_backend),
 ) -> Table:
     return backend.optimization.tables.create(**table.model_dump())
+
+
+@autodoc
+@router.delete("/{table_id}/")
+def delete(
+    table_id: int,
+    backend: Backend = Depends(deps.get_backend),
```
meksor, line 93:
```suggestion
@router.delete("/{id}/")
def delete(
    id: int,
    backend: Backend = Depends(deps.get_backend),
```

ixmp4/server/rest/optimization/variable.py
```diff
@@ -85,3 +85,12 @@ def create(
     backend: Backend = Depends(deps.get_backend),
 ) -> OptimizationVariable:
     return backend.optimization.variables.create(**variable.model_dump())
+
+
+@autodoc
+@router.delete("/{variable_id}/")
+def delete(
+    variable_id: int,
+    backend: Backend = Depends(deps.get_backend),
```
meksor, line 94:
```suggestion
@router.delete("/{id}/")
def delete(
    id: int,
    backend: Backend = Depends(deps.get_backend),
```

tests/core/test_optimization_equation.py
```diff
@@ -91,10 +91,48 @@ def test_create_equation(self, platform: ixmp4.Platform) -> None:
...
+        run.optimization.equations.delete(item=equation.name)
+
+        assert run.optimization.equations.tabulate().empty
+
+        # Confirm that IndexSet has not been deleted
+        assert not run.optimization.indexsets.tabulate().empty
+
+        # Test that association table rows are deleted
+        # If they haven't, this would raise DeletionPrevented
+        run.optimization.indexsets.delete(item=indexset_1.id)
```
meksor, line 134: Please add a check making sure all the associated data is also deleted, for example the entries in the association table.

meksor (changes requested):

tests/core/test_optimization_equation.py
glatterf42: Please clarify this request: line 134 does test that rows in the indexset/equation association table are deleted (though indirectly). Do you think this indirect test is not enough? 
Or is this about the indexset/indexsetdata association table? I'd hope we have a test for that in `test_delete_indexset`.

And I have to admit: I'm not sure how to test this short of somehow grabbing the session and executing a `SELECT` for some association table. Is this what you have in mind?

meksor (approved):

PR #164, opened 2025-02-12
Make `Equation.indexsets` optional

glatterf42:
Working on the ixmp4 shim, I discovered that our assumption that `Equation`s must always be constrained to `IndexSet`s is wrong: in message_ix, there are [numerous examples of `Equation`s not linked to `IndexSet`s](https://github.com/iiasa/message_ix/blob/main/message_ix/models.py#L680-L712).

This PR includes some minor improvements to the code, but mainly makes the `Equation.indexsets` attribute optional.

### TODO

- ~[ ] I'm not sure the changes here need their own DB migration. It looks like only an ORM attribute/property is affected, but make sure that's true; otherwise, include the migrations.~
  `alembic check` tells me there are no new upgrade operations for this PR :)

### DIFF
--- a/ixmp4/data/db/optimization/equation/model.py
+++ b/ixmp4/data/db/optimization/equation/model.py
@@ -65,8 +65,9 @@ class Equation(base.BaseModel):
     )
 
     @property
-    def indexset_names(self) -> list[str]:
-        return [indexset.name for indexset in self._indexsets]
+    def indexset_names(self) -> list[str] | None:
+        names = [indexset.name for indexset in self._indexsets]
+        return names if bool(names) else None
 
     @property
     def column_names(self) -> list[str] | None:
--- a/tests/core/test_optimization_equation.py
+++ b/tests/core/test_optimization_equation.py
@@ -40,33 +40,55 @@ class TestCoreEquation:
     def test_create_equation(self, platform: ixmp4.Platform) -> None:
         run = platform.runs.create("Model", "Scenario")
 
-        # Test normal creation
-        indexset_1, indexset_2 = tuple(
+        # Test creation without indexset
+        equation_1 = run.optimization.equations.create("Equation 1")
+        assert equation_1.run_id == run.id
+        assert equation_1.name == "Equation 1"
+        assert equation_1.data == {}
+        assert equation_1.indexset_names is None
+        assert equation_1.column_names is None
+        assert equation_1.levels == []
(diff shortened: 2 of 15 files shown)


ixmp4/data/db/optimization/equation/model.py
```diff
@@ -65,8 +65,9 @@ def validate_data(self, key: Any, data: dict[str, Any]) -> dict[str, Any]:
     )
 
     @property
-    def indexset_names(self) -> list[str]:
-        return [indexset.name for indexset in self._indexsets]
+    def indexset_names(self) -> list[str] | None:
+        names = [indexset.name for indexset in self._indexsets]
+        return names if bool(names) else None
```
glatterf42, line 70: Note that this is different from the Variable property:

```python
    def indexset_names(self) -> list[str] | None:
        return (
            [indexset.name for indexset in self._indexsets]
            if len(self._indexsets) > 0
            else None
        )
```

I don't remember for sure why I wrote this differently in the first place. I think at some point I ran into a `indexset has no attribute 'name'` error, but I don't see how this could happen. Please let me know if you do, then I should change the equation property back again. 
If this truly can't happen, I should probably change the variable definition for consistency.

meksor (approved):

tests/core/test_optimization_equation.py
```diff
@@ -40,35 +40,57 @@ class TestCoreEquation:
     def test_create_equation(self, platform: ixmp4.Platform) -> None:
```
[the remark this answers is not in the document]
glatterf42: I think this criticism applies to more tests than this (keeping in mind also your feedback on the other PRs), so let's do this properly: in a subsequent PR. For now, I'll open an issue to keep this in mind.

ixmp4/data/db/optimization/equation/model.py
glatterf42: Either implementation works fine in for both item types (i.e. clears the test suite). The only difference I can see is that `bool([None]) == True`, so with the current equation implementation, we might get `[None]` back if `self._indexsets = [None]`. However, I think this is not really possible anymore; I think it was at some point, but now, if an Equation or Variable doesn't have any IndexSets, `self._indexsets` will simply be `[]` (an empty list since no `_Association` table entry will be created). So I would keep the current equation implementation and use it for variables, too, since I find it more intelligible.

PR #166, opened 2025-03-04
Change Log, Checkpoints and Run-Locks

meksor:
This PR is meant to contain the three major parts of what we internally often call the "versioning" feature.

## Change Log

The change log is an in-database representation of all transactions run on a given ixmp4 platform. 
This requires the generation of "version tables" from database models (by adding columns and removing constraints) and a central "transaction" table tracking all transactions committed, referenced by each version table.
This part of the feature is logically/structurally built around the concepts in a database, so it might actually cover more functionality than needed (f.e. there is no way to update a region, but the system tracks update statements on the region table).
Currently this is handled by `sqlalchemy-history` as it has support for alembic migrations and can be replaced with `sqlalchemy-continuum` which would enable database-native versioning for postgres via postgres triggers.
Enabling and disabling these features on a per-platform basis (to avoid performance problems with sqlite) is still a little unclear to me and will need a little more time investment. 

**The tricky parts:**
1. **This change entails an inevitable performance regression.** While it will not change scaling characteristics (think big O), it will worsen all linear operations siginificantly. What was once a single query statement might become three or more (sequential ones!) with this change:

```sql
INSERT (a, b, c) INTO xy;

# becomes:

INSERT (a, b, c) INTO xy;
INSERT (datetime.now()) INTO transactions;
INSERT (a, b, c, transaction__id, operation_type) INTO xy_versions;
```
Most of the overhead can be alleviated by using database-native table triggers, but these are hard to write and maintain so I would suggest relying on a library for that as well

2. **What to version & handling relationships**

Versioning of the data points table puts me in front of a choice:
They can either:
a) Prevent deletes entirely for all regions, variables and units - hiding the unused ones when `SELECT`ing
or
b) Version every single table in the database that is subject to a relation and query the version variant of the related table when rolling back

I went with option b).

## Run Checkpoints

To be able to distinguish different workflow steps in a sea of "transaction" records (one ixmp4 api call might create multiple records) we will need to add another table to ixmp4 which references **run__id** and **transaction__id** and stores a user-supplied **message**. 

We now have /four/ relevant tables.

The original table **xy**.

| a | b | c |
| -  |  - |  - |
| A | B | C |

The version table **xy_version**.

| a | b | c | transaction_id | end_transaction_id | operation_type |
| -  |  - |  - |  - |  - |  - |
| A | B | C | 1 | NULL | 0 |

The transaction table **transactions**.

| id | issued_at |
| -  |  - | 
| 1 | 13:37 27.08.1999 |

The new **checkpoints** table.
*May also include a timestamp and user information.*

| message | transaction_id | run_id |
| - | - | - | 
| "Hello World" | 1 | 1 |

The last table (checkpoints) can be used to query all workflow steps for a specific run and to revert to said steps by looking at which transaction record they point to and replaying or rolling back the change history.
This would probably look kind of like this: 

```py
run = ...
with run.acquire_lock():
    run.checkpoints.create("Create run")
    run.iamc.add(my_data)
    run.checkpoints.create("Add data")

run.checkpoints.tabulate()

# msg | transaction_id
# "Create run" | 3
# "Add data" | 5

run.checkpoints.list()[0].revert()

run.iamc.tabulate()

# empty ... 
```

## Run Locks 

To make sure that a user does not include unintended changes by a different user, a locking mechanism is needed for each "run" object/row. A user shall be able to acquire a lock, preventing all other users from writing to objects related to the run.  
Still unclear to me:
If a user tries to acquire the lock to a locked run, should the operation:
 a) block until the other user releases the lock
 b) fail immediately

Also, I remember the uploads crashing and not releasing the lock being a problem in the past. How do we want to handle this in the future?


glatterf42:
@khaeru, this PR and the open questions are likely interesting to you, too. Please chime in if you know the answers/have some preferences for those :)

I imagine people would not be happy to wait an unknown amount of time on someone else releasing a lock, so my guess is: if you need a lock but can't get it, we should immediately raise an error.
And regarding the failed uploads in the past: I'm not sure what our options are. Theoretically, we could release any lock once an operation finishes, successful or not, but if it's not successful, additional edits from some other person who doesn't know about the failure might not be helpful. Is there a way to mark the `Run` as "broken" somehow, so that whoever acquires the lock next knows that they're getting into something? Maybe they could even reach out to whoever had the lock last before them.

meksor:
> The PR is adding tables with a column 'transaction_id' and also the name (in the example code snippet) 'checkpoint', which I sense are the same thing.

No! Checkpoints connect a particular run (-id) to a transaction, they serve as points to roll back to, but i think this might be irrelevant for the message integration anyway.

> I don't know if that makes it possible to have a different implementation, perhaps one with less performance impact (time, or storage, e.g. by not storing more than ~2 transaction_id associated with a given run_id) or maintenance overhead, but providing that info in case.

The revert logic is always the same: select the relevant objects from the version table at the particular transaction, then delete all current data and insert the old data. For the checkpoint we get `run__id` and `transaction__id` from the checkpoint itself, for the `run.transact()` feature we just have to remember the transaction at which the run was locked (currently done in a new column in the run table).

> Similar but distinct, we did have frequent issues where users would call ixmp.TimeSeries.check_out() (usually as message_ix.Scenario.check_out(), which is identical). Then their code would error or crash. Because they never explicitly called TimeSeries.commit() or TimeSeries.discard(), the database would record that the object (~ixmp4 Run) was "checked out to" them, and not allow any user (them, or another) to change it.

> In other words, if only "atomic transactions" via with run.acquire_lock() are provided, that would be livable. We can add noisy warnings to ixmp.TimeSeries.check_out() during the shim/migration period that say check_out() will not ever be supported with IXMP4Backend and encourage users to switch to using acquire_lock/transact. The 'check_out'/…/'discard' workflow was I guess designed with an eye to interactive use, e.g. in IPython/Jupyter, but with .transact() we can now teach users "best practices" without the code being more verbose.

Ok good, i think this approach is preferable for me as well.

Example cases: 1 and 2 seems doable with the same logic, but that logic is out of scope for this PR. 
I think, theoretically, the chance of the race condition you describe is there, but very small. 

3.: Okay, the timeout needs to be implemented, will need to look into that. 

> It would be good to consider whether words like 'transact(ion)', 'commit', 'checkpoint', etc. have existing usage and meanings (a) in ixmp, message_ix etc. or (b) in likely users' mind ('commit' e.g. because many people may also be Git users; fewer are RDBMS users); and consider whether the meaning in ixmp4/message_ix is different. This can help us avoid inviting potential confusion, as that creates work to steer users clear or help correct their misunderstanding.

I agree, unfortunately all the names i can think of are taken. PSQL even has "checkpoints" as well, so in fact all these terms ('transact(ion)', 'commit', 'checkpoint') have (sometimes multiple) other meanings.

meksor:
Hi, the first version of this feature is available on this branch.
There are still a few open issues which I might try to solve on auxiliary branches and merge into this one.

- **exchange sqlachemy-history for sqlachemy-continuum** in order to use its sql trigger generation functionality. This will (hopefully, presumably) get rid of some of the manual version object generation code (f.e. in `bulk_insert_versions`) and speed up this PR so the regression is acceptable.
- **Disable versioning for specific platforms** as the sql trigger generation will only work for postgres. This is tricky since I assume I need to find a way to run the sqlalchemy mapper in a "reentrant" way. It also means throwing and testing a new exception.
- **Implement versioning and revert logic for optimization data**, which is best done by fridolin or at least with fridolins help.

Further I need clarification on some issues by @danielhuppmann and @khaeru. Details follow.

## Regions and Units

As of now regions and units are tracked as versioned objects, but not rolled back with a run. This means, the following code will throw a `RegionNotFound` error.

```py
region1 = platgorm.regions.create("Region 1")
platgorm.regions.create("Region 2")

run = platform.runs.create(model, scenario)
data_with_region1 = pd.DataFrame(...) # Has IAMC Data that references "Region 1"
data_with_region2 = pd.DataFrame(...) # Has IAMC Data that references "Region 2"

with run.transact():
    run.iamc.add(data_with_region1)
    run.checkpoints.create("Add data with region 1")

    run.iamc.add(data_with_region2)
    run.checkpoints.create("Add data with region 2")

    run.iamc.remove(data_with_region1)
    run.checkpoints.create("Remove data with region 1")

platform.regions.delete(region1)

with run.transact():
    # Revert to "Add data with region 1"
    run.checkpoints[1].revert() # Will throw an error!!

```

This seems good to me, because if incorrect regions were deleted, we do not want to implicitly create them again when rolling back a run.
What do you think?

## Automatic Rollback

Right now, every time a run exits the `transact()` context manager, it will roll back to the last checkpoint created. This means, if the user forgets to create a checkpoint after adding data, the run will be empty.
This adds quite a bit of overhead, especially via the web API.

```py
with run.transact():
    run.iamc.add(test_data_1)
    run.checkpoints.create("Add test data 1")
    run.iamc.add(test_data_2)

# the run will now contain only test_data_1
assert run.iamc.tabulate() == test_data_1
```

Is that right or should the data stay?

## Client-Side Locking Mechanism

The run lock is now implemented in the facade layer using the two backend methods `backend.run.lock` and `backend.run.unlock`. The client has to keep track of which locks it owns as the methods behave as follows:

`backend.run.lock(run_id)` will check the specified run's `lock_transaction` column. If the column contains `NULL`, it will be updated to contain the id of the last transaction in the `transactions` table.
If the column already contains a `transaction.id`, the `Run.IsLocked` exception is thrown.

`backend.run.unlock(run_id)` will always set the `lock_transaction` column to `NULL`, regardless of the current value.

Hence, the `transact` contextmanager logic tries to aquire the lock via `backend.run.lock(run_id)` and tracks that the operation succeeded in the `owns_lock` property of the corresponding run facade model.

```py

    @contextmanager
    def transact(self) -> Generator[None, None, None]:
        self._model = self.backend.runs.lock(self._model.id)
        self.owns_lock = True

        yield

        checkpoint_df = self.checkpoints.tabulate()
        checkpoint_transaction = int(checkpoint_df["transaction__id"].max())

        assert self._model.lock_transaction is not None

        if checkpoint_transaction > self._model.lock_transaction:
            self.backend.runs.revert(self._model.id, checkpoint_transaction)
        else:
            self.backend.runs.revert(self._model.id, self._model.lock_transaction)

        self._model = self.backend.runs.unlock(self._model.id)
        self.owns_lock = False
```

A more sophisticated and secure locking mechanism is a lot more work, as it requires me to keep track of the users that have locked runs. Posponed for an indefinite time, alright?

## LockRequired Exception

Since the client now keeps track of which locks it owns, we can throw exceptions in the facade layer when an operation is attempted that requires a lock.
Currently this applies to:

- `run.iamc.add`
- `run.iamc.remove`
- `run.checkpoints.create()`

A list of operations in the optimization part will have to be worked out as well, can you think of anything else? Did I understand the assignment correctly?

## Run Meta Indicators/Data

Are not covered by the rollback logic or the LockRequired exception, should this be the case?

I'm sure as soon as I post this comment I will think of something else, but for now: please advise!

khaeru:
> Regions and Units
> 
> As of now regions and units are tracked as versioned objects, but not rolled back with a run.
> […]
> What do you think?

As mentioned, message_ix/current ixmp test suites and usage don't need this, so no opinion.

> Automatic Rollback
> […]
> Is that right or should the data stay?

Stay? This is what I meant above about "following semantics of the standard library". See [How to use the connection context manager](https://docs.python.org/3/library/sqlite3.html#how-to-use-the-connection-context-manager) in the docs for the `sqlite3` module. This is the behaviour that the current [`ixmp.TimeSeries.transact()`](https://docs.messageix.org/projects/ixmp/en/stable/api.html#ixmp.TimeSeries.transact) is trying to emulate—see the example there in the docs.

> Client-Side Locking Mechanism
> […]
> A more sophisticated and secure locking mechanism is a lot more work, as it requires me to keep track of the users that have locked runs. Postponed for an indefinite time, alright?

Yes, that's fine. If something gets inadvertently locked by another user (and it shouldn't if we're encouraging users to use the context manager), it's already a great improvement that `backend.run.unlock()` will be accessible, instead of having to do something like [this](https://iiasa-ece.slack.com/archives/C03M5NX9X0D/p1675155635233089?thread_ts=1675153202.295639&cid=C03M5NX9X0D).

> LockRequired Exception
> […]
> A list of operations in the optimization part will have to be worked out as well, can you think of anything else? Did I understand the assignment correctly?

This is more for @glatterf42 I think. Just to clarify, one point above was that creating a Run instance *at all* should be safe against collisions. That is to say, rather than locking a existing Run while certain things are done to it, two simultaneous attempts to create new (not-yet-existing) Runs that want the same or similar identifiers should be safe.

If that's already achieved some other way than the locking mechanism you've written, e.g. by lower-level transaction control in the DB layer, then we're fine.

> Run Meta Indicators/Data
> […]
> Are not covered by the rollback logic or the LockRequired exception, should this be the case?

I'm not aware of people using these through `ixmp`/`message_ix`, so no opinion.

danielhuppmann:
Thanks @meksor, this is very nice!

> Regions and Units
> 
> As of now regions and units are tracked as versioned objects, but not rolled back with a run.
> […]
> What do you think?

This makes sense.

> Automatic Rollback
> […]
> Is that right or should the data stay?

Would it make sense that the `transact()`-context ends with setting a checkpoint, and the checkpoint-description is part of the transact-arguments?

> Client-Side Locking Mechanism
> […]
> A more sophisticated and secure locking mechanism is a lot more work, as it requires me to keep track of the users that have locked runs. Postponed for an indefinite time, alright?

Agree with @khaeru, that's ok for our use case.

> LockRequired Exception and Run Meta Indicators/Data
> […]
> Are not covered by the rollback logic or the LockRequired exception, should this be the case?

Please add creation and deletion of meta indicators to the operations that are change-logged and require a lock.

Adding the optimization-items to the change-log-and-locks-logic can be left for a follow-up PR (led by @glatterf42), in my opinion.

meksor:
> > Automatic Rollback
> > […]
> > Is that right or should the data stay?
> 
> Would it make sense that the `transact()`-context ends with setting a checkpoint, and the checkpoint-description is part of the transact-arguments?

Yes, sounds good. The rollback now only happens upon an exception.

> > LockRequired Exception and Run Meta Indicators/Data
> > […]
> > Are not covered by the rollback logic or the LockRequired exception, should this be the case?
> 
> Please add creation and deletion of meta indicators to the operations that are change-logged and require a lock.

Understood.

meksor:
Ok, a few updates on the points from above: 

> - exchange sqlachemy-history for sqlachemy-continuum in order to use its sql trigger generation functionality. This will (hopefully, presumably) get rid of some of the manual version object generation code (f.e. in bulk_insert_versions) and speed up this PR so the regression is acceptable.

I did this and it worked like a drop-in replacement, no changes besides adjusting the imports necessary.
However this does not really matter because I could not get the next objective to work:

> - Disable versioning for specific platforms as the sql trigger generation will only work for postgres. This is tricky since I assume I need to find a way to run the sqlalchemy mapper in a "reentrant" way. It also means throwing and testing a new exception.

Unfortunately the libraries I used here are not made for this and enabling/disabling versioning is only possible globally with our current code-base. This also means that I cannot use the "native versioning" feature from `sqla_continuum` because the trigger setup will fail every time on sqlite databases. 

Making all of this work will take a lot more effort and might mean that I have to implement large parts of the library myself. (In which case I will try to get rid of the dependency entirely)
More specifically: I need to find a way to run the sqlalchemy mapper when a backend/platform is created as opposed to how its done currently (when the classes are imported). This means rebuilding the large parts of the Model/Repository code in the data layer. Its probably wise to consider a bigger refactor if im getting my hands dirty with something like that already. 

**What does this mean for this PR?**
I suggest finishing up here by making sure the tests cover the desired use-cases and living with the performance regression (pretty much exactly **2x** on my setup). The big thing missing here are the meta-indicators afaics. 
This allows fridolin to continue with his part in the optimization modules.
Next, Ill try to focus on a few bug-fixes and the `scse-toolkit` integration since that has been on my back-log for ages. Ill have to make do without the test-container registry (which would be nice for this change) as ICT told us that they are "going to ignore all my tickets for 2 months". Hopefully that also fixes any issues that @danielhuppmann has encountered in the meantime. 

Then, I will try to reduce the amount of code in this project while keeping the versioning performance improvement in mind. Right now we have some complicated inheritance hierarchies, a lot of boilerplate code and not a lot of internal documentation which can all be improved. A big improvement here would be to combine "abstract" and "db" layers (since the interface is pretty much completely defined now) and to generate the API server and client code with less copy-pasta and boilerplate-o-rama. Getting rid of the filter classes so we can use the `TypedDict`s introduced in the mypy PR as the sole source-of-truth for the possible arguments is also on my list. 

After that Ill look at reworking this feature, at that point it should be a lot more straight forward (i might eat my words later, but ill stay optimistic for now).
This is a few months in the future. OK @danielhuppmann ?

_Side note_: For some reason **one single CI test** across all the jobs is failing:
[py3.10 | backend=sqlite,rest-sqlite | with-pyarrow=true | pgsql=16 | pandas=2.1.0](https://github.com/iiasa/ixmp4/actions/runs/14358969557/job/40259369584?pr=166#logs)
the test in question is "test_mp_tabulate_big[rest-sqlite]". Ive been trying to reproduce this locally for a few days now with no luck (I installed the relevant deps and python versions, tried `gh act pull_request`, the test locally always pass). Has anyone ever encountered something like this before? @glatterf42 @pmussak @phackstock

meksor:
Revert logic and LockRequired exceptions are implemented for the meta indicators now, FYI.

pmussak:
> _Side note_: For some reason **one single CI test** across all the jobs is failing: [py3.10 | backend=sqlite,rest-sqlite | with-pyarrow=true | pgsql=16 | pandas=2.1.0](https://github.com/iiasa/ixmp4/actions/runs/14358969557/job/40259369584?pr=166#logs) the test in question is "test_mp_tabulate_big[rest-sqlite]". Ive been trying to reproduce this locally for a few days now with no luck (I installed the relevant deps and python versions, tried `gh act pull_request`, the test locally always pass). Has anyone ever encountered something like this before? @glatterf42 @pmussak @phackstock

I did not encounter this yet, but I looked into it

ixmp4/core/run.py
```diff
@@ -73,6 +81,42 @@ def unset_as_default(self) -> None:
...
+        except Exception as e:
+            checkpoint_df = self.checkpoints.tabulate()
+            checkpoint_transaction = int(checkpoint_df["transaction__id"].max())
+
+            assert self._model.lock_transaction is not None
+
+            if checkpoint_transaction > self._model.lock_transaction:
+                self.backend.runs.revert(self._model.id, checkpoint_transaction)
+            else:
+                self.backend.runs.revert(self._model.id, self._model.lock_transaction)
```
glatterf42, line 111: We could cover this line with the existing tests with something like

```python
transaction = checkpoint_transaction if checkpoint_transaction > self._model.lock_transaction else self._model.lock_transaction

self.backend.runs.revert(self._model.id, transaction)
```

ixmp4/data/abstract/checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+class EnumerateKwargs(
+    annotations.HasRunIdFilter,
+    annotations.HasTransactionIdFilter,
+    total=False,
+):
+    pass
+
+
+class Checkpoint(base.BaseModel, Protocol):
+    """ """
```
glatterf42, line 23: Why is this empty docstring required?

ixmp4/data/abstract/checkpoint.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class CheckpointRepository(
+    base.Creator,
+    base.Deleter,
+    base.Enumerator,
+    Protocol,
+):
+    def create(self, run__id: int, message: str) -> Checkpoint:
+        """Creates a checkpoint.
```
glatterf42, line 45:
```suggestion
        """Creates a Checkpoint.
```

To distinguish from the regular word.

glatterf42: I'd do the same for 'run', 'model', and 'checkpoint' below.

ixmp4/data/abstract/meta.py
```diff
@@ -37,9 +37,14 @@ def from_pytype(cls, type_: type) -> str:
     "Key for the entry. Unique for each `run__id`."
     dtype: types.String
     "Datatype of the entry's value."
-    value: types.Integer | types.Float | types.Integer | types.Boolean
+    value: types.Integer | types.Float | types.String | types.Boolean
```
glatterf42, line 40: Oh, thanks for catching that :)

ixmp4/data/abstract/run.py
glatterf42, line 1: I would again capitalize words that represent objects in our code.

ixmp4/data/api/base.py
```diff
@@ -91,7 +91,8 @@ def df_to_dict(df: pd.DataFrame) -> DataFrameDict:
         index=df.index.to_list(),
         columns=columns,
         dtypes=dtypes,
-        data=df.values.tolist(),
+        # https://github.com/numpy/numpy/issues/27944
+        data=df.values.tolist(),  # type: ignore
```
glatterf42, line 95: Apparently, this was not necessary before. I accept that we now need it, but could you please then specify the error code for mypy (e.g. `type: ignore[assignment]`)? This would document and clarify the issue for future reference :)

ixmp4/data/api/checkpoint.py
```diff
@@ -0,0 +1,61 @@
+from typing import TYPE_CHECKING, Any, ClassVar, cast
+
+if TYPE_CHECKING:
+    pass
```
glatterf42, line 4: Why import `TYPE_CHECKING` if we don't need it?

ixmp4/data/db/checkpoint.py
```diff
@@ -0,0 +1,93 @@
+import pandas as pd
+from typing_extensions import Unpack
```
glatterf42, line 2: Let's hope we don't forget this line when dropping support for Python 3.11 :)
Though come to think of it, maybe this is what you meant by 'internal documentation': we could have one document somewhere listing these things once instead of one extra comment line for every import.

ixmp4/data/db/iamc/datapoint/repository.py
```diff
@@ -1,10 +1,12 @@
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 import numpy as np
 import pandas as pd
 import pandera as pa
 from pandera.engines import pandas_engine
 from pandera.typing import DataFrame, Series
+from sqlalchemy_continuum import version_class
```
glatterf42, line 8: Didn't we want to restrict all sqlalchemy imports to `db.__init__.py` or so? Does that not apply to related imports such as this or was this never a strict rule?

ixmp4/data/db/iamc/datapoint/repository.py
```diff
@@ -204,7 +207,7 @@ def tabulate(
 
     def check_df_access(self, df: pd.DataFrame) -> None:
         if self.backend.auth_context is not None:
-            ts_ids = set(df["time_series__id"].unique().tolist())
+            ts_ids = cast(set[int], set(df["time_series__id"].unique().tolist()))
```
glatterf42, line 210: I've run into a few errors by now that originated from calling `cast()` when the underlying object could actually be something else, too. This example seems fine to me, but pandas (not sure if Series or DataFrames) also have this `.astype()` function which you could use instead to convert to `int` explicitly without needing `cast()`, I think :)

ixmp4/data/db/iamc/timeseries/repository.py
```diff
@@ -103,7 +104,7 @@ def tabulate(self, **kwargs: Unpack[EnumerateKwargs]) -> pd.DataFrame:
     @guard("edit")
     def bulk_upsert(self, df: pd.DataFrame, create_related: bool = False) -> None:
         if self.backend.auth_context is not None:
-            run_ids = set(df["run__id"].unique().tolist())
+            run_ids = cast(set[int], set(df["run__id"].unique().tolist()))
```
glatterf42, line 107: Similar to above with `astype()`.

glatterf42: (There are a few more in similar files below.)

ixmp4/data/db/optimization/indexset/model.py
```diff
@@ -27,7 +27,7 @@ def data(self) -> list[float] | list[int] | list[str]:
         return (
             []
             if self._data_type is None
-            else np.array([d.value for d in self._data], dtype=self._data_type).tolist()
+            else np.array([d.value for d in self._data], dtype=self._data_type).tolist()  # type: ignore
```
glatterf42, line 30: Please add the specific error type again :)

ixmp4/data/db/run/repository.py
```diff
@@ -173,16 +174,135 @@ def set_as_default_version(self, id: int) -> None:
...
+            suffixes=(None, "_y"),
+        )
+
+    def revert_iamc_data(self, run: Run, transaction__id: int) -> None:
+        current_dps = self.backend.iamc.datapoints.tabulate(
+            run={"id": run.id, "default_only": False}
+        )
+
+        if not current_dps.empty:
+            self.backend.iamc.datapoints.bulk_delete(current_dps)  # type: ignore[arg-type]
```
glatterf42, line 214: Yes, perfect :)

ixmp4/data/db/run/repository.py
```diff
@@ -173,16 +174,135 @@ def set_as_default_version(self, id: int) -> None:
...
+        )
+        self.backend.iamc.datapoints.bulk_upsert(version_dps)  # type: ignore[arg-type]
+
+    def revert_meta(self, run: Run, transaction__id: int) -> None:
+        current_meta = self.backend.meta.tabulate(
+            run={"id": run.id, "default_only": False}
+        ).drop(columns=["id", "value", "dtype"])
+
+        if not current_meta.empty:
+            self.backend.meta.bulk_delete(current_meta)  # type: ignore[arg-type]
```
glatterf42, line 242: But now I'm wondering why this file needs so many `# type: ignore`s. How much work would it be to get rid of them?

ixmp4/db/migrations/env.py
```diff
@@ -1,7 +1,24 @@
+"""
+When making schema migrations for versioned tables
```
glatterf42, line 2: How does this integrate with the existing workflow (using `alembic revision --autogenerate`)?

ixmp4/server/rest/checkpoint.py
```diff
@@ -0,0 +1,55 @@
...
+            offset=pagination.offset,
+            table=bool(table),
+        ),
+        total=backend.units.count(_filter=filter),
+        pagination=pagination,
+    )
+
+
+@router.post("/", response_model=api.Checkpoint)
+def create(
```
glatterf42, line 43: This method is also exposed in the facade layer, if I'm not mistaken, so why not `@autodoc` it here?

glatterf42: Reading the other diffs, I'm now wondering in general how we choose which functions to decorate like this.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ toml = ">=0.10.2"
 psycopg = { extras = ["binary"], version = ">=3.1.12" }
 pydantic-settings = ">=2.1.0"
 rich = ">=13.5.2"
+sqlalchemy-continuum = "^1.4.2"
```
glatterf42, line 37:
```suggestion
sqlalchemy-continuum = ">=1.4.2"
```
For consistency.

pyproject.toml
```diff
@@ -82,12 +83,15 @@ build-backend = "poetry_dynamic_versioning.backend"
 requires = ["poetry-core>=1.2.0", "poetry-dynamic-versioning"]
 
 [tool.mypy]
-exclude = [
-    '^ixmp4\/db\/migrations\/',
-]
-disable_error_code = ['override']
+exclude = ['^ixmp4\/db\/migrations\/']
+disable_error_code = ['override', 'unused-ignore']
```
glatterf42, line 87: Why do we need the `unused-ignore` here? Isn't that just indicating that a `type: ignore` can be removed, and thus always useful and easy to handle when it pops up?

tests/core/test_meta.py
```diff
@@ -13,10 +14,12 @@ def test_run_meta(platform: ixmp4.Platform) -> None:
     run1.set_as_default()
 
     # set and update different types of meta indicators
-    # NOTE mypy doesn't support setters taking a different type than their property
-    # https://github.com/python/mypy/issues/3004
-    run1.meta = {"mint": 13, "mfloat": 0.0, "mstr": "foo"}  # type: ignore[assignment]
-    run1.meta["mfloat"] = -1.9
+    # NOTE mypy doesn't support setters taking a different type than
```
glatterf42, line 17: I would probably move this inside the context manager block, but it can also stay here if you prefer.

ixmp4/data/db/base.py
```diff
@@ -524,13 +686,96 @@ def bulk_delete_chunk(self, df: pd.DataFrame) -> None:
...
+        super().__init__(backend)
+
+        self.version_bundle: Bundle[Any] = Bundle(
+            self.version_class.__name__,
+            *db.utils.get_columns(self.version_class).values(),
+        )
+
+    def select_transactions(self) -> db.sql.Select[Any]:
+        exc = db.select(self.transaction_class)
+        return exc
```
glatterf42, line 724: What's the difference between this and simply

```python
    return db.select(self.transaction_class)
```

glatterf42 (approved):
Wow, great work! Some questions and remarks inline, but generally speaking, this looks good to me :)
I may still need your help to adjust the optimization part and other PRs to this.

ixmp4/core/run.py
meksor: Yeah, but that would be a trick and not actually provide additional coverage for the functionality. I can add a test that covers the line in question properly.

ixmp4/data/abstract/checkpoint.py
meksor: We dont capitalize these in all other parts of the codebase (at least the parts I wrote). I would argue that it makes sense to capitalize if we actually use a link to the class like `:class:\`pandas.Series\``, but in that case its probably better to change all occurances of this, not just the ones added in this PR

ixmp4/data/abstract/run.py
meksor: I would again push back on that for now

ixmp4/data/api/base.py
meksor: Sure, to be clear I think this has something to do with the pandas version used.

ixmp4/data/db/base.py
meksor: As you can see, there is no difference :) 
This is left over from when this function was more complex

ixmp4/core/run.py
glatterf42: That would be great, thanks :)

ixmp4/data/db/iamc/datapoint/repository.py
glatterf42: But `cast()` only makes sense when we can be sure we're casting to the correct type, right? Otherwise, I could cast `"foo"` as a `pd.DataFrame` and mypy wouldn't complain if I access `"foo".dtypes`, though the code would obviously still fail. So if we're sure that `df["time_series__id"]` is a `Series[int]`, does `.astype()` then actually change that or does it have some fast check in place to make it a noop in that case? Their source code looks like that to me, studying it for a minute.
But of course, we don't need to make this change, just wanted to raise this point: if we ever change the type of `df["time_series__id"]`, `cast()` will still happily mark it as `Series[int]`, even if it isn't that anymore.

ixmp4/server/rest/checkpoint.py
meksor: autodoc literally just adds the functions module locator string thingy to the docstring so the rest endpoint documentation links to the specific functions. Its not that important and I had bigger plans for it as far as I remember which were never acted upon or bad in the first place. I added it to the endpoints I saw were missing it for consistency

pyproject.toml
meksor: So I added all these ignores because I had a numpy version installed which needed them, but if one has a different version installed mypy will instead complain about these "unused ignores", hence I added this option.

tests/core/test_meta.py
meksor: makes sense, done

pyproject.toml
phackstock: Those are two different requirements. Using the caret `^1.4.2` means this: `1.4.2<=version<2.0`, while `>=1.4.2` means that there's no upper limit for the version. I'd keep the caret.

meksor: Yes, but sqlalchemy is also specified like this, so I'd say its best to pin both the one or the other way.

phackstock:
> Side note: For some reason one single CI test across all the jobs is failing:
[py3.10 | backend=sqlite,rest-sqlite | with-pyarrow=true | pgsql=16 | pandas=2.1.0](https://github.com/iiasa/ixmp4/actions/runs/14358969557/job/40259369584?pr=166#logs)
the test in question is "test_mp_tabulate_big[rest-sqlite]". Ive been trying to reproduce this locally for a few days now with no luck (I installed the relevant deps and python versions, tried gh act pull_request, the test locally always pass). Has anyone ever encountered something like this before? @glatterf42 @pmussak @phackstock

I haven't seen this exact error but I've definitely encountered some pandas python interaction weirdness along those lines. Like some particular pandas version not playing nice with an older version of python. In the end the solution was usually to bump up the minimum version of pandas and the problem went away.

ixmp4/core/run.py
meksor: Added a test, line is now covered

ixmp4/data/abstract/checkpoint.py
meksor: Probably just forgot about it. I added a simple docstring.

meksor:
> Is there a special reason why you're testing python 3.10 with pandas 2.1.0 and then for 3.11 bump it up to 2.1.3?

The reason is that we want to test these particular pandas versions explicitly since they've been problematic in the past, but also don't want to necessarily add another 3 tests for each of them to cut down on time waiting for the jobs to finish.

phackstock:
@meksor, ah I see, thanks for the clarification. You could try running the tests on GitHub with python 3.10 and pandas 2.1.3 to see if those pass. If they do maybe increment the minimum version of pandas to 2.1.3

meksor:
I fixed it by truncating a few lines of the test data, so i am guessing that its some sort of memory problem where:
 - python 3.10 is more memory hungry
 - this PR doubles (?) memory consumption for some operations
 - the API platform needs more memory for JSON serialization

phackstock:
oh interesting

pyproject.toml
glatterf42: Using `">="` was required by @danielhuppmann once in order for ixmp4 to support as many older versions as possible.

PR #167, opened 2025-03-04
Expose `list_docs()` in the facade & enable `list_docs(id__in=...)`

glatterf42:
Here's the PR you asked for, @danielhuppmann: it exposes the `list()` functionality we already have for docs for most facade objects and enables a filter so that only IDs from a provided list are considered.

There are several things to consider, though:

1. I don't know if this is how you envisioned the facade API (i.e. providing IDs; the other docs functions often translate one name to one ID first, which would need to be adapted for a list of names).
2. The PR comes with tests, but these are not exhaustive. For instance, I don't enforce anywhere that only one of `dimension_id` or `dimension_id__in` can be used. In fact, if both are used, we silently discard the latter at the moment. This behavior is not tested.
3. The code includes quite a bit of repetition, which is not great style/potentially even technical debt. However, I didn't see how else to do this unless I'm rewriting more significant parts of the docs feature, which might be worthwhile itself, but not the target here.
4. Currently, the optimization items in the facade layer all have a `.docs` property through which one can manage each item's docs, but their repository classes lack any docs functionality. So rather than adding all of that here, I assumed we don't urgently need this and thus did not expose the `list_docs()` function for them.

### DIFF
--- a/ixmp4/data/db/docs.py
+++ b/ixmp4/data/db/docs.py
@@ -1,3 +1,4 @@
+import logging
 from collections.abc import Iterable
 from typing import ClassVar, TypeVar
 
@@ -11,6 +12,8 @@ from ixmp4.data import abstract, types
 
 from ..auth.decorators import guard
 from . import base
+
+logger = logging.getLogger(__name__)
 
 
 class AbstractDocs(base.BaseModel):
@@ -70,7 +73,13 @@ class BaseDocsRepository(
         # TODO Should we enforce that only one of these filters can be active?
         if dimension_id is not None:
             _exc = _exc.where(self.model_class.dimension__id == dimension_id)
-        elif dimension_id__in is not None:
+        if dimension_id__in is not None:
+            if dimension_id and dimension_id not in dimension_id__in:
+                logger.warning(
+                    "Applying incompatible filters to select for docs: "
+                    f"dimension_id '{dimension_id}' is not in dimension_id__in "
+                    f"{dimension_id__in}!"
+                )
             _exc = _exc.where(self.model_class.dimension__id.in_(dimension_id__in))
 
         return _exc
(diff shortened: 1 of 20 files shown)


ixmp4/data/db/docs.py
```diff
@@ -60,20 +62,29 @@ def select(
...
         dimension_id: int | None = None,
+        dimension_id__in: Iterable[int] | None = None,
     ) -> db.sql.Select[tuple[DocsType]]:
         if _exc is None:
             _exc = db.select(self.model_class)
 
+        # TODO Should we enforce that only one of these filters can be active?
         if dimension_id is not None:
             _exc = _exc.where(self.model_class.dimension__id == dimension_id)
+        elif dimension_id__in is not None:
```
meksor, line 73:
```suggestion
        if dimension_id__in is not None:
```
This might seem kind of unnecessary, but it makes sure that the parameter is not just quietly ignored when `dimension_id` is supplied.

ixmp4/data/db/docs.py
```diff
@@ -60,20 +62,29 @@ def select(
...
     def select_for_count(
-        self, _exc: db.sql.Select[tuple[int]], dimension_id: int | None = None
+        self,
+        _exc: db.sql.Select[tuple[int]],
+        dimension_id: int | None = None,
+        dimension_id__in: Iterable[int] | None = None,
     ) -> db.sql.Select[tuple[int]]:
         if dimension_id is not None:
             _exc = _exc.where(self.model_class.dimension__id == dimension_id)
+        elif dimension_id__in is not None:
```
meksor, line 86:
```suggestion
        if dimension_id__in is not None:
```
ditto

ixmp4/data/db/unit/docs.py
```diff
@@ -17,5 +18,12 @@ class UnitDocsRepository(
     dimension_model_class = Unit
 
     @guard("view")
-    def list(self, *, dimension_id: int | None = None) -> list[AbstractDocs]:
-        return super().list(dimension_id=dimension_id)
+    def list(
+        self,
+        *,
+        dimension_id: int | None = None,
+        dimension_id__in: Iterable[int] | None = None,
```
meksor, line 25: I am unsure about the use of `Iterable` in this PR. On one hand, we use it all over the code base, especially in `TypedDicts`, on the other hand its not a super robust or helpful type hint:
```py
>>> from typing import Iterable 
>>> isinstance([], Iterable)
True
>>> isinstance({}, Iterable)
True
```
I seem to also faintly remember having a problem with pydantic models where `Iterables` would be empty once they are consumed a single time, not sure if im remembering that correctly. 

I think it might be worth thinking about switching to `list[xyz]` for most of the things marked `Iterable` currently.

meksor (approved):
Looks good, two comments to consider, may require follow up PR.

ixmp4/data/db/unit/docs.py
glatterf42: Safed in a follow-up issue :)

PR #169, opened 2025-03-27
Handle adding empty optimization data

glatterf42:
In message_ix, we may iterate over all variables etc of one `Scenario` (backed by a `Run`) to add them to another `Scenario` (backed by another `Run`), which may include adding empty data (`{}` or `pd.DataFrame()`) to them. This should be working already as only the combination of existing and new data are validated. 
However, this led to an error for indexsets because adding an empty list violates one of the `UniqueConstraint`s for their underlying tables. So I was adjusting that already and figured the remaining items would also benefit from tests covering this case.
On top of which, I figure we don't need to bother the DB or validate data when adding empty data. This PR avoids that, which should make the code more performant. 

NOTE: I'm also using this branch right now for some temporary debug commits.


meksor (approved):
lgtm

PR #171, opened 2025-04-04
Remove data from linked items when removing data from indexset items

glatterf42:
`message_ix.Scenario.remove_set()` expects that when an element from an `IndexSet` is removed, all corresponding keys from other optimization items (i.e. keys that contain this element from items linked the `IndexSet`) are also removed (because otherwise, the linked items would contain invalid data). This PR adds the functionality to provide the same behavior.

However, since I'm not a big of silently doing stuff in the background, our version of this comes with a flag to disable it. The default for this is `True` right now, i.e. to remove related data, but this is mainly because the function will for now mostly be called from message_ix, which needs this default. I'm happy to swap the default to `False` and have the shim layer request `True` every time it needs it if you think that's better.

The PR includes some preparatory cleanups and notes.

### DIFF
--- a/ixmp4/data/db/optimization/indexset/repository.py
+++ b/ixmp4/data/db/optimization/indexset/repository.py
@@ -5,19 +5,16 @@ from typing import TYPE_CHECKING, Any, List, Literal, Sequence, cast
 from typing_extensions import Unpack
 
 from ixmp4.data.db.optimization.equation.model import (
-    Equation,
     EquationIndexsetAssociation,
 )
 from ixmp4.data.db.optimization.equation.repository import EquationRepository
 from ixmp4.data.db.optimization.parameter.model import (
-    Parameter,
     ParameterIndexsetAssociation,
 )
 from ixmp4.data.db.optimization.parameter.repository import ParameterRepository
-from ixmp4.data.db.optimization.table.model import Table, TableIndexsetAssociation
+from ixmp4.data.db.optimization.table.model import TableIndexsetAssociation
 from ixmp4.data.db.optimization.table.repository import TableRepository
 from ixmp4.data.db.optimization.variable.model import (
-    OptimizationVariable,
     VariableIndexsetAssociation,
 )
 from ixmp4.data.db.optimization.variable.repository import VariableRepository
@@ -33,7 +30,7 @@ from ixmp4 import db
 from ixmp4.data.abstract import optimization as abstract
 from ixmp4.data.auth.decorators import guard
 
-from .. import base
+from .. import base, utils
 from .docs import IndexSetDocsRepository
 from .model import IndexSet, IndexSetData
 
@@ -56,6 +53,25 @@ class IndexSetRepository(
         from .filter import OptimizationIndexSetFilter
 
         self.filter_class = OptimizationIndexSetFilter
+
+        self._linked_columns_lookup = {
+            "table": (
+                TableIndexsetAssociation.table__id,
+                TableIndexsetAssociation.indexset__id,
+            ),
+            "parameter": (
+                ParameterIndexsetAssociation.parameter__id,
+                ParameterIndexsetAssociation.indexset__id,
+            ),
+            "equation": (
+                EquationIndexsetAssociation.equation__id,
+                EquationIndexsetAssociation.indexset__id,
+            ),
+            "variable": (
+                VariableIndexsetAssociation.variable__id,
+                VariableIndexsetAssociation.indexset__id,
+            ),
+        }
 
     def add(self, run_id: int, name: str) -> IndexSet:
         indexset = IndexSet(run__id=run_id, name=name)
@@ -138,7 +154,7 @@ class IndexSetRepository(
         _data = [str(d) for d in data] if isinstance(data, list) else [str(data)]
 
         if remove_dependent_data:
-            REPOS: dict[
+            repos: dict[
                 Literal["table", "parameter", "equation", "variable"],
                 TableRepository
                 | ParameterRepository
@@ -150,11 +166,9 @@ class IndexSetRepository(
                 "equation": self.backend.optimization.equations,
                 "variable": self.backend.optimization.variables,
             }
-            for kind, ids in find_all_linked_item_ids(
-                session=self.session, indexset_id=id
-            ):
-                remove_invalid_data(
-                    repo=REPOS[kind],
+            for kind, ids in self.find_all_linked_item_ids(id=id):
+                self.remove_invalid_linked_data(
+                    repo=repos[kind],
                     ids=ids,
                     indexset_name=self.get_by_id(id=id).name,
                     data=_data,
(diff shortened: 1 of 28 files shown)


tests/core/test_optimization_indexset.py
```diff
@@ -136,17 +137,104 @@ def test_remove_elements(self, platform: ixmp4.Platform) -> None:
...
+            for k, v in parameter.data.items()
+        }
+        expected_parameter_2 = {
+            k: [
+                v[i]
+                for i in range(len(v))
+                if parameter_2.data[indexset_1.name][i] not in remove_data
+            ]
+            for k, v in parameter_2.data.items()
+        }
```
meksor, line 216: Once again, I think its better if these tests are more explicit and simple. This test is currently 130 lines long, hard to read and generates its own "truths"...

tests/core/test_optimization_indexset.py
```diff
@@ -136,17 +137,104 @@ def test_remove_elements(self, platform: ixmp4.Platform) -> None:
...
         indexset_1.remove(data=remove_data)
 
         assert indexset_1.data == expected
 
-        # Test removing single item
-        expected.remove("fa")
-        indexset_1.remove(data="fa")
+        # TODO Why do we need to reload items when using the rest API?
+        if isinstance(platform.backend, RestBackend):
```
meksor, line 222: This is a no-go. The tests should work for all types of platforms the exact same way. Either /always/ reload the objects on all platforms to show that this is intended and the user is required to do this or fix the reloading in the facade or api layers somehow.

ixmp4/data/db/optimization/indexset/repository.py
```diff
@@ -130,3 +175,156 @@ def remove_data(
...
+
+
+def find_all_linked_item_ids(
+    session: db.Session, indexset_id: int
+) -> Generator[
+    tuple[Literal["table", "parameter", "equation", "variable"], Sequence[int]],
+    Any,
+    None,
+]:
+    """Finds all optimization items in `session` linked to an IndexSet.
```
meksor, line 236: Hm... This implies that this function uses the sqla-orm session object registry to find items, which is not the case. 
Is this docstring/function AI generated?

tests/core/test_optimization_indexset.py
meksor: ditto the other tests...

meksor (changes requested):

ixmp4/data/db/optimization/indexset/repository.py
```diff
@@ -108,13 +127,39 @@ def add_data(
...
+                TableRepository
+                | ParameterRepository
+                | EquationRepository
+                | VariableRepository,
+            ] = {
+                "table": self.backend.optimization.tables,
+                "parameter": self.backend.optimization.parameters,
+                "equation": self.backend.optimization.equations,
+                "variable": self.backend.optimization.variables,
+            }
```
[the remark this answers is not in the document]
glatterf42: Just making sure: if I do that, we depend on `IndexSetRepository` being initialized _after_ the repos specified here. In `__init__`, that is currently not the case because the entries in e.g. `backend/db/SqlAlchemyBackend.make_repositories()` are sorted alphabetically. Do you want me to change that, too, presumably including a NOTE that this order is crucial, or should we keep the `repos` variable inside this function (which will only be called after all repos have been initialized)?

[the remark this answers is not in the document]
meksor: Hm, thats true and even worse... Two options:
1. Leave as is
2. Instantiate new repositories as properties of this repository in `__init__` as it's done in other repos as well...
(This is where it bites us that we didnt properly split up service and repository layers as usual with this design pattern)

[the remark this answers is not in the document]
glatterf42: Then I will leave it as is for now :)
But could you please remind me of the name of the design pattern? I haven't come across the terms 'service layer' and 'repository layer' and I want to read up on them :)

ixmp4/data/db/optimization/indexset/repository.py
glatterf42: Nothing here is AI-generated, but apparently, my own understanding wasn't good enough to produce a good docstring. I thought that using `session.scalars()` meant I was using the sqla-orm session object.

ixmp4/data/db/optimization/indexset/repository.py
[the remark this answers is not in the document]
meksor: Sometimes its called MRS (Model, Repository, Service) Pattern. Sometimes just Repository-Service Pattern. https://pmichaels.net/service-repository-pattern/

tests/core/test_optimization_indexset.py
glatterf42: I'm not sure I agree with this, unfortunately. Yes, the test is long, but this saves me from writing a lot of boilerplate because for each of these (thinking of each Parameter/Table object as a separate case), the setup and eventual comparison are very similar. Of course, I could avoid that with a fixture immediately setting up some Indexsets on a Run, but this would still run multiple times for the various test functions we would then have.
In contrast, we now set up the Indexsets once and then also set up various items that will be affected by `indexsets.remove_data()`. I think this is simply a consequence of electing to have `indexsets.remove_data()` secretly affect these items: in a sense, this turns `test_remove_data()` into both unit and integration test. Though maybe I can split at least these two cases, then.

As for the truth generation, I started hardcoding the `expected_parameter` values. But then I would also need to hardcode `expected_table` and `expected` for consistency, and I'm not sure that's our goal here. `expected` reads `[data for data in test_data if data not in remove_data]`, which is just using readily understandable Python logic and there's nothing unexpected that could happen here: `test_data` and `remove_data` are clearly defined above.
One might be tempted to say `parameter.data` is not as clearly defined, but I think it is: we have our own tests ensuring this is exactly what we expect based on our input. So in that sense, `expected_parameter` is also safe to use, not changing unexpectedly unless the `test_parameter_add_data` also fails. 
So of course, I can hardcode our expectations here; I just don't think it's necessary. I'll do it so you can take a look and see if you like it better :)

ixmp4/data/db/optimization/indexset/repository.py
```diff
@@ -130,3 +175,156 @@ def remove_data(
...
+        EquationIndexsetAssociation.indexset__id,
+    ),
+    "variable": (
+        VariableIndexsetAssociation.variable__id,
+        VariableIndexsetAssociation.indexset__id,
+    ),
+}
+
+
+def _find_linked_item_ids(
```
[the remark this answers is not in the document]
meksor: > wrote these functions outside of the Repo simply because they didn't need to be linked
Well... But they are all using the session which is kind of why we have the repository classes in the first place, I think thats reason enough to keep them there...

> This means that the IndexSetRepo will now have a slightly different API, but I guess that's fine. Now users could call IndexSetRepo.remove_invalid_linked_data() with potentially ill-suited input. Our test suite should guard against this leading to mistakes, though.

Yeah, unfortunately we dont have real encapsulation in python (public, protected, private) except for some very hacky-feeling double-underscore behaviour:

```py
class Test(object):
    def __private(self):
        print("foo")

Test()._Test__private()
> "foo"
```

This is generally present across the code-base, but since `Backend` should be hinted with the abstract layer interface those methods should not show up in the IDE when not using the concrete decendants directly.

> I really don't understand the second point, though: I am already using the various other repositories to delete the data, am I not? We find all linked items through a query and then use their item type to access the corresponding repo (with repo.list(id__in=ids) and call repo.remove_data().

Well so each repository would have `remove_invalid_data` and `find_linked_item_ids` methods that just pertain to their model type. This version is kind of dodging the fact that there is no proper inheritance hierarchy for all the functionality that is shared between `OptimizationItems`. A symptom of this is that each model class has its own `indexset_names` and `column_names` property for example...

tests/core/test_optimization_indexset.py
glatterf42: I agree. Always reloading manually is an option, but feels like a waste: after all, it seems that sqlalchemy immediately updates the session objects and reflects that new `.data` without the need for a manual reload. So I would prefer to make the api layer reload automatically, but I don't see a way yet.

glatterf42: There are two kinds of issues here: we have these conditional reloads in the core tests and the data tests for different reasons.
In the data tests, it seems that `add_data()` changes the object in the session, but this is not reflected in our API layer. I think this can be resolved by having `add_data()` return the object/model data, just like `create()`. This would change our API to 

```python
indexset = platform.backend.optimization.indexsets.create().add_data()
```

This should be fine on the core side, too, since we currently do

```python
        self.backend.optimization.indexsets.add_data(id=self._model.id, data=data)
        self._model = self.backend.optimization.indexsets.get(
            run_id=self._model.run__id, name=self._model.name
        )
```

which would simply become

```python
        self._model = self.backend.optimization.indexsets.add_data(id=self._model.id, data=data)
```

The other part seems trickier: the core tests currently manually reload the linked items when running via the restAPI (for `test_remove_data()`, which affects more items). These linked items are also automatically expired and reloaded by sqla when connecting directly to the DB layer, but I'm not sure how to best reflect this in the API layer.

ixmp4/data/db/optimization/indexset/repository.py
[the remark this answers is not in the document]
glatterf42: Thanks for the clarification :)
I would prefer to stay with the current setup. Each repo already has a `remove_data()` function and it seems kind of unnecessary to me have give each a new `remove_invalid_data()` function, which is just a special case of the former. The current setup makes sense to me because the IndexSetRepo knows which values have become invalid and so can easily call the existing `remove_data()` function in the right way.

tests/core/test_optimization_indexset.py
meksor: > Yes, the test is long, but this saves me from writing a lot of boilerplate because for each of these (thinking of each Parameter/Table object as a separate case), the setup and eventual comparison are very similar. 

You can use shared, well-named functions for this purpose.

> but this would still run multiple times for the various test functions we would then have.

So? Saving 3 seconds of test execution time is less important than having isolated, easily assessable tests in my view.

>  In contrast, we now set up the Indexsets once and then also set up various items that will be affected by indexsets.remove_data(). I think this is simply a consequence of electing to have indexsets.remove_data() secretly affect these items: in a sense, this turns test_remove_data() into both unit and integration test. Though maybe I can split at least these two cases, then.

Hm, not sure I would call that an integration test... But yes, splitting the tests "simple testing of delete functionality" and "testing if the delete functionality affects the other db-tables as well" up already adds value if the latter fails. 

> As for the truth generation, I started hardcoding the expected_parameter values. But then I would also need to hardcode expected_table and expected for consistency, and I'm not sure that's our goal here. expected reads [data for data in test_data if data not in remove_data], which is just using readily understandable Python logic and there's nothing unexpected that could happen here: test_data and remove_data are clearly defined above.

You are right, the first line is pretty easy to read, however: 
  1. You are relying on some piece of code here that you "think" will produce the right output. Now, I am sure that it is actually correct, but the point is that you would need to step through with a debugger to /really/ make sure. In this case its obviously a bit "ad absurdum" but the point still stands. Also, just hardcoding this stuff is even shorter that what we currently have, it seems unnecessarily complicated for no reason (other than that if you change the test case a bit you need to actually /think/ about what this test does to do it properly, i dont think thats a bad thing).
  2. The other lines are even more complicated and shaky, `expected_parameter` f.e. could use `enumerate` instead of this combination of `range` and `len`, and again, the hardcoded version would be the same length and easier to read. The only argument youre giving me is "it cant change anyway" and thats not even what im worried about. I just want to be able to read these very simple tests in a simple way, so that if I (or you) have to fix something in a year, we wont have to wade through a swamp of doubly-nested list comprehensions to get there...

tests/core/test_optimization_indexset.py
meksor: Implementing an sqla-like object tracking system is too much work I think. Just dont do it at all, the user should be aware that the data they have might be stale. Not sure how ixmp currently handles this? Probably not at all?

tests/core/test_optimization_indexset.py
glatterf42: Okay, thanks for clarifying our priorities here :)
I'll improve the tests in the way you describe, but I'd like to do this in a subsequent PR as this has also come up in a different review (see #176) and thus I'd like to go through the test suite with sufficient time to make sure we apply these adjustment to all optimization tests that need them.

ixmp4/data/db/optimization/indexset/repository.py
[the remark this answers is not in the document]
meksor: Ok then please at least move it to inside this repository.

tests/core/test_optimization_indexset.py
meksor: Sounds good.

ixmp4/data/db/optimization/indexset/repository.py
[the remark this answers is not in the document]
glatterf42: This is already done :)

tests/data/test_optimization_equation.py
```diff
@@ -526,3 +596,32 @@ def test_tabulate_equation(self, platform: ixmp4.Platform) -> None:
...
+        # the same order
+        equation_2 = platform.backend.optimization.equations.create(
+            run_id=run.id,
+            name="Equation 2",
+            constrained_to_indexsets=equation.indexset_names,
+            column_names=equation.column_names,
+        )
+
+        assert equation_2.column_names == column_names
+        assert equation_2.indexset_names == indexset_names
```
glatterf42, line 627: I just noticed that for #157, I was thinking of also including tests like this: ensuring that what @meksor described actually works: one item is created with a certain order of `column_names`, another is created based on that ordering and should store the identical ordering. 
Do you think we need those? If so, I can also add them for variables, parameters, and tables, and for the core layer.

tests/core/test_optimization_indexset.py
glatterf42: I agree that this would likely take us too far. Especially since you're right: ixmp doesn't worry about such things because it never returns actual objects called "IndexSet" or so. 
`init_set()` simply [returns `None`](https://github.com/iiasa/ixmp/blob/main/ixmp/backend/base.py#L857), so whenever you want to access an IndexSet's data with ixmp, you do `set()`, which loads the data and [returns a `pandas.Series`](https://github.com/iiasa/ixmp/blob/main/ixmp/backend/base.py#L916) in an IndexSet's case.
So in a sense, ixmp always requires manual reloading. We're already offering more comfort by providing these objects, and if one operates directly via the DB layer, even `remove_data()` is directly reflected in all stored objects. Since this would be hard to replicate on the API layer, though, we should clearly document that users are always expected to manually reload stored objects if they are affected by an `Indexset.remove_data()` operation. 
We could also include a footnote that this "might not be necessary" when working directly with the DB layer, but only in a way that makes it clear that this is just something users who know what they're doing should use.

So for the test suite, all these objects are manually reloaded now on both backends :)

tests/data/test_optimization_equation.py
meksor: I would add a testcase for `run.clone()` that tries to clone items with both `column_names` and `indexset_names`
as that is currently untested i think. I don't think its necessary here.

meksor: But also fine at a later date if it seems to work (also seeing as there is a precautionary `order_by` parameter passed now)

glatterf42: Okay, thanks, dropping this commit.

PR #173, opened 2025-04-24
Add `timeout` parameter to `Run.transact()`

meksor:
Like the title says, this PR adds a timeout parameter to the newly added `Run.transact` context manager. 
The docstring I added pretty much sums up what it does: 

```py
    @contextmanager
    def transact(
        self, message: str, timeout: float | None = None
    ) -> Generator[None, None, None]:
        """
        Context manager to lock the run before yielding control
        back to the caller. The run is unlocked and a checkpoint
        with the provided `message` is created after the context
        manager exits.
        If an exception occurs, the run is reverted to the last
        checkpoint or if no checkpoint exists, to the transaction
        the run was locked at.

        If the run is already locked, the context manager will
        throw `Run.IsLocked` or if `timeout` is provided retry until
        the timeout has passed (and then throw the original
        `Run.IsLocked` exception).

        Parameters
        ----------
        messsage : str
            The message for the checkpoint created after
            conclusion of the context manager.
        timeout : int, optional
            Timeout in seconds.

        Raises
        ------
        :class:`ixmp4.core.exceptions.RunIsLocked`
            If the run is already locked and no timeout is provided
            or the provided timeout is exceeded.
        """
```

The timeout is computed as follows for each iteration (no maximum for the amount of iterations is set):

```py
sleep_time = elapsed_time * 2
sleep_time = min(sleep_time, self.maximum_lock_timeout)
sleep_time = max(sleep_time, self.minimum_lock_timeout)
sleep_time = min(sleep_time, remaining_time)
```
Where `maximum_lock_timeout = 5.` and `minimum_lock_timeout = .1`.

This means for a timeout of `1` the logic will try to aquire the lock at the following times after the call to `transact`:
- 0 s (immediately)
- 0.1 s (minimum_lock_timeout)
- 0.3 s (0.1 + (0.1*2))
- 0.9 s (0.3 + (0.3*2))
- 1 s (`timeout`)

Does this fit your requirements @khaeru @glatterf42 ?

### DIFF
--- a/ixmp4/core/run.py
+++ b/ixmp4/core/run.py
@@ -1,3 +1,4 @@
+import time
 from collections import UserDict
 from contextlib import contextmanager
 from typing import ClassVar, Generator, cast
@@ -36,7 +37,9 @@ class Run(BaseModelFacade):
 
     checkpoints: RunCheckpoints
 
-    owns_lock = False
+    owns_lock: bool = False
+    minimum_lock_timeout: float = 0.1
+    maximum_lock_timeout: float = 5
 
     def __init__(self, **kwargs: Unpack[RunKwargs]) -> None:
         super().__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


glatterf42 (approved):
This looks good to me, though I can't say if this mimics the existing behavior in ixmp or if the `timeout` there works somehow differently.

EDIT: I don't think the current `ixmp.timeseries.transact()` has a timeout parameter. I can imagine testing this frequently is fine, if anything, I understand @khaeru's original

> `run.acquire_lock()` blocks for a specific, configurable amount of time to wait for User D's operation to conclude.

such that checking only once (at the end of the specified timeout) would suffice.

PR #174, opened 2025-04-28
Add `Run.delete()` and `Backend.runs.delete()`

### LINKED ISSUE #29: Add feature to delete a run
We need to add a feature to delete a run, and there are two strategies for implementation.

- A "true" deletion: delete all entries from the RUN table, iamc datapoints, measurands, meta indicators.
- Hiding a run for non-admin users ("soft delete"): this would be a nice option to allow modelers (non-admins) to remove a run but still have the option to easily restore it

The main rationale for this distinction is that I do not want to allow non-admin users to hard-delete runs uploaded to a Scenario Explorer instance - the current practice is that the modelers have to send me an email and I decide if the deletion is ok (e.g. wrong scenario name -> ok; some error in some timeseries data -> just upload a new version).

meksor:
Adds backend and facade layer functionality for the deletion of a run. 
Deleting a run also deletes all the associated data, use with care. Users may only delete a run with "manage" permissions set. 
~~Note that while `Run.delete()` requires locking the run, `Platform.runs.delete()` does not. More info in #29 
This test illustrates the behaviour:~~
Edit: this has changed, more info below.

Note also that deleting a run **deletes all checkpoints for the run as well** since the checkpoint table is not versioned, this will **not be recoverable**. 
Adds the endpoint `DELETE run/{id}/`.

Closes #29


meksor:
Coveragebot will complain about the new abstract method until #142 is merged.

meksor:
Of course I found a bug in the versioning logic in this PR, I fixed it here as well. (Not filtering the versioned rows correctly by `run__id`, inserting the past versions associated with **all runs** by mistake)

ixmp4/core/run.py
```diff
@@ -99,6 +99,9 @@ def transact(self, message: str) -> Generator[None, None, None]:
 
         try:
             yield
+        # except Forbidden as e:
+        #     self._unlock()
+        #     raise e
```
glatterf42, line 104: Why are we keeping these lines?

ixmp4/data/db/run/repository.py
```diff
@@ -84,6 +85,38 @@ def add(self, model_name: str, scenario_name: str) -> Run:
         self.session.add(run)
         return run
 
+    def delete_optimization_data(self, id: int) -> None:
+        pass  # TODO: Implement this. @glatterf42
```
glatterf42, line 89: Noted, but this will have to wait for merging all PRs required for passing the message tutorials and releasing ixmp and message_ix (with all the cleanups this entails).

ixmp4/data/db/run/repository.py
```diff
@@ -84,6 +85,38 @@ def add(self, model_name: str, scenario_name: str) -> Run:
...
+
+    def delete_meta_data(self, id: int) -> None:
+        current_meta = self.backend.meta.tabulate(
+            run={"id": id, "default_only": False}
+        ).drop(columns=["id", "value", "dtype"])
+
+        if not current_meta.empty:
+            self.backend.meta.bulk_delete(current_meta)  # type: ignore[arg-type]
```
glatterf42, line 97: We should fix this discrepancy with pydantic and pandas DataFrames at some point.

ixmp4/data/db/run/repository.py
```diff
@@ -232,6 +265,9 @@ def revert_iamc_data(self, run: Run, transaction__id: int) -> None:
         )
         self.backend.iamc.datapoints.bulk_upsert(version_dps)  # type: ignore[arg-type]
 
+    def revert_optimization_data(self, run: Run, transaction__id: int) -> None:
+        pass  # TODO: Implement this. @glatterf42
```
glatterf42, line 269: Noted, but same as above.

glatterf42 (approved):
Thanks, looks good to me :)

ixmp4/core/run.py
meksor: we forgot to take out the trash :D

ixmp4/data/db/run/repository.py
meksor: sure, i wanted to provide these stubs so its not forgotten

ixmp4/data/db/run/repository.py
meksor: agreed

meksor:
Update: As suggested by @danielhuppmann, the delete backend method now automatically tries to acquire the run lock. 
If the user or another (second) user already locked the run, the call will fail. This makes things a lot easier in both the context manager logic and from a consistency standpoint. 
Tests to illustrate behaviour: 
```py

    def test_run_delete_locked_run(self, platform: ixmp4.Platform) -> None:
        self.small.load_dataset(platform)
        run1_1 = platform.runs.get("Model 1", "Scenario 1")
        run1_2 = platform.runs.get("Model 1", "Scenario 1")

        with run1_1.transact(""):
            with pytest.raises(RunIsLocked):
                run1_1.delete()
            with pytest.raises(RunIsLocked):
                run1_2.delete()

    def test_run_delete_via_object_method(self, platform: ixmp4.Platform) -> None:
        self.small.load_dataset(platform)
        run1 = platform.runs.get("Model 1", "Scenario 1")
        run2 = platform.runs.get("Model 2", "Scenario 2")

        for run in [run1, run2]:
            run.delete()
            self.assert_run_data_deleted(platform, run)

    def test_run_delete_via_repository_id(self, platform: ixmp4.Platform) -> None:
        self.small.load_dataset(platform)
        run1 = platform.runs.get("Model 1", "Scenario 1")
        run2 = platform.runs.get("Model 2", "Scenario 2")

        for run in [run1, run2]:
            platform.runs.delete(run.id)
            self.assert_run_data_deleted(platform, run)

    def test_run_delete_via_repository_object(self, platform: ixmp4.Platform) -> None:
        self.small.load_dataset(platform)
        run1 = platform.runs.get("Model 1", "Scenario 1")
        run2 = platform.runs.get("Model 2", "Scenario 2")

        for run in [run1, run2]:
            platform.runs.delete(run)
            self.assert_run_data_deleted(platform, run)

```

PR #182, opened 2025-05-27
Add ID filters to /iamc/datapoints endpoint

meksor:
Currently the `/iamc/datapoints` endpoint does not allow filtering by the IDs of related rows. This PR adds the ID field for the filter classes in question. 

Closes #180

PR #183, opened 2025-06-20
Maintenance June 2025

glatterf42:
This PR addresses a few things that I saved for later in recent weeks: 

1. It fixes an outdated poetry instruction: the managed environment can no longer be activated as before with poetry 2.0, instead [we need something like](https://python-poetry.org/docs/managing-environments/#activating-the-environment) `eval $(poetry env activate)`.
2. Mypy 1.16 has brought ~1050 errors in 34 files to light. These can almost all be fixed by adjusting our abstract data models: their attributes need to be annotated as types directly instead of as `Mapped[<type>]`.
3. The mypy sqlalchemy plugin is deprecated and removed in this PR (as mentioned [here](https://github.com/iiasa/ixmp4/pull/182#issuecomment-2975242886)).
4. Various dependencies are bumped to their latest available versions, even removing some version pins (this is largely experimental, lets see if the tests like this change).

Finally, for mypy 1.16, we can now have [property setters and getters with different types](https://mypy.readthedocs.io/en/stable/changelog.html#different-property-getter-and-setter-types) :)
I remember that this was a hurdle somewhere in ixmp4, but I can't find where exactly, so my workaround is blending in well, it seems. Should we need this in the future, though, (or still uncover it with this PR), we can also use this feature now :)


ixmp4/db/filters.py
```diff
@@ -320,7 +322,7 @@ def apply(
         session: db.Session,
     ) -> db.sql.Select[tuple[FilterType]]:
         dict_model = dict(self)
-        for name, field_info in self.model_fields.items():
+        for name, field_info in self.__class__.model_fields.items():
```
glatterf42, line 325: This resolves the DeprecationWarning mentioned [here](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_fields).

glatterf42:
Just noticed a few things:

- I forgot a Warning about [alembic's changing configuration keys](https://alembic.sqlalchemy.org/en/latest/changelog.html#change-5a2673f270038185b4ee24c3cee0cb13). Also note that they may switch to [using pyproject.toml for their config at some point, too](https://alembic.sqlalchemy.org/en/latest/tutorial.html#using-pep-621), but for the moment, `alembic.ini` is still the default option.
- I didn't know that we had flakiness like [this](https://github.com/iiasa/ixmp4/actions/runs/15830548216/job/44621881421) in our test suite. I haven't investigated this in detail yet, but I'm not immediately sure why we have this at all.
- Lastly, there are a few more DeprecationWarnings on Python 3.13 runs of the sort:

  ```
  tests/core/test_iamc.py::TestCoreIamc::test_run_annual_datapoints_raw[sqlite]
  /home/runner/work/ixmp4/ixmp4/.venv/lib/python3.13/site-packages/pandas/core/flags.py:53: ResourceWarning: unclosed database in <sqlite3.Connection object at 0x7f0fbb2a73d0>
    self._obj = weakref.ref(obj)
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
  ```
  Here, I again don't know where this is coming from. These warnings existed before this PR, though.

PR #184, opened 2025-06-26
Require Run lock for certain "optimization" operations

glatterf42:
The "optimization" part still needs versioning, change log and Run locking as the rest of the code got in #166. This PR begins this process by adding `run.require_lock()` to the operations where I think it makes sense: `create()`, `delete()`, `add_data()`, and `remove_data()` (or their equivalents).

It also re-enables mypy's warnings about unused `type: ignore` statements as I think this is useful and works well with the latest versions of mypy and numpy. This immediately produced 24 errors in 9 files, all of which I resolved again.

Please note that this PR may look huge, but actually isn't. The vast majority of changed lines originates from the test suite, where many lines had to be adjusted to suit the `with run.transact()` context manager in formatting only, so you can likely skip all `test_` files when reviewing :)

### DIFF
--- a/ixmp4/core/iamc/data.py
+++ b/ixmp4/core/iamc/data.py
@@ -60,7 +60,7 @@ class RunIamcData(BaseFacade):
 
     def add(self, df: pd.DataFrame, type: DataPointModel.Type | None = None) -> None:
         self.run.require_lock()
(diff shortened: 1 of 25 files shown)


meksor (changes requested):
It might be wise to add tests that check if the RunLockRequired exception is raised. As of now, existing tests are adjusted to account for the change but the change isn't actually tested itself.

PR #186, opened 2025-07-09
pin pandera to >=0.24.0 

meksor:
We use the new way of importing from pandera so we need to pin the version to at least the version that introduced it.


phackstock (approved):
Perfect, thanks for the quick fix.

PR #187, opened 2025-07-29
Enable vaing mixed IAMC datatpoints in a Run

danielhuppmann:
This PR enables adding different `DataPoint`-types to a **Run** (one type after the other using separate calls of `Run.iamc.add()`.

Turns out that the having mixed datapoint-types and querying them already worked implicitly.

Request for specific feedback:
1. With some tinkering, I found that simply adding 
   ```python
    existing_df.dropna(axis=1, inplace=True)
    ```
    is sufficient to enable adding datapoints of a type to a Run that already has datapoints of another type. Without that additional line, this raises an error. However, with that line removed, the `meta` tests fail. Any better idea?
2. I extended the workhorse of the iamc-datapoints-tests `do_run_datapoints()` to split the dataframe by type if necessary. This could be simplified to always split by type (avoiding four if-else statements), but that may have other side effects.
3. I didn't test this yet with categorical data.

### DIFF
--- a/ixmp4/data/db/base.py
+++ b/ixmp4/data/db/base.py
@@ -421,6 +421,7 @@ class BulkOperator(Tabulator[ModelType]):
     ) -> pd.DataFrame:
         columns = db.utils.get_columns(self.model_class)
         primary_key_columns = db.utils.get_pk_columns(self.model_class)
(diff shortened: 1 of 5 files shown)


meksor:
This should be fixed in #219

PR #195, opened 2025-08-19
Sync In-Situ Adjustment to Initial Versions Migration

meksor:
This PR makes sure to tell the database driver not to load all result rows at once so the migrated table has no natural size limit due to memory constraints.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -133,7 +133,7 @@ jobs:
       #  install root project
       #------------------------
       - name: Install library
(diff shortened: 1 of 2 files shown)


meksor:
`scenariocompass-dev` grew to about 30GB and does not fit into my laptops memory. I originally used `partition()` only, which i thought would do this in the first place, but it turns out partition only partitions the python list, the database driver (psycopg here for example) still potentially loads the whole query at once. I guess the last few weeks pushed the database over my laptops memory constraints even without the python overhead.

PR #196, opened 2025-08-19
api.Run.created at nullability

meksor:
There is a mistake in the `api.Run` model, namely that `created_by` is nullable in the database but not that model resulting in errors.


glatterf42 (approved):
The PR and branch name seem confusing since we also have the `created_at` attribute, but only `created_by` is changed here. But again, once tests pass, lgtm :)

meksor:
sorry, i seem to have been sloppy...

Issue #198, opened 2025-08-28
test data generator fails - missing run lock


pmussak:
While running https://github.com/iiasa/ixmp4-ts I just found that the test data generator in ixmp4 which can be run with

`ixmp4 platforms generate test --datapoints 100`

fails with `RunLockRequired`
<details>

<summary>stacktrace</summary>

```
...
Generating Datapoint(s)... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:--
╭──────────────────────────────────────────────────────────────────── Traceback (most recent call last) ────────────────────────────────────────────────────────────────────╮
│ /home/pino/projects/secondary/ixmp4/ixmp4/cli/platforms.py:306 in generate                                                                                                │
│                                                                                                                                                                           │
│   303 │   │   │   num_units,                                                                                                                                              │
│   304 │   │   │   num_datapoints,                                                                                                                                         │
│   305 │   │   )                                                                                                                                                           │
│ ❱ 306 │   │   generate_data(generator)                                                                                                                                    │
│   307 │   │   utils.good("Done!")                                                                                                                                         │
│   308                                                                                                                                                                     │
│   309                                                                                                                                                                     │
│                                                                                                                                                                           │
│ ╭────────────────────────────────────── locals ──────────────────────────────────────╮                                                                                    │
│ │      generator = <ixmp4.data.generator.MockDataGenerator object at 0x7cbd42fa8440> │                                                                                    │
│ │          lines = [                                                                 │                                                                                    │
│ │                  │   ' - \x1b[36m10\x1b[0m Model(s) ',                             │                                                                                    │
│ │                  │   ' - \x1b[36m40\x1b[0m Run(s) ',                               │                                                                                    │
│ │                  │   ' - \x1b[36m200\x1b[0m Region(s) ',                           │                                                                                    │
│ │                  │   ' - \x1b[36m1000\x1b[0m Variable(s) ',                        │                                                                                    │
│ │                  │   ' - \x1b[36m40\x1b[0m Units(s) ',                             │                                                                                    │
│ │                  │   ' - \x1b[36m100\x1b[0m Datapoint(s) '                         │                                                                                    │
│ │                  ]                                                                 │                                                                                    │
│ │           name = 'Datapoint(s)'                                                    │                                                                                    │
│ │ num_datapoints = 100                                                               │                                                                                    │
│ │     num_models = 10                                                                │                                                                                    │
[...]

PR #199, opened 2025-08-28
use new run lock before adding iamc data in data generator

pmussak:
Addresses #198 by wrapping `run-iamc.add` in `with run.transact`.

### DIFF
--- a/ixmp4/data/generator.py
+++ b/ixmp4/data/generator.py
@@ -105,7 +105,8 @@ class MockDataGenerator(object):
             df["region"] = region_name
             df["variable"] = variable_name
             df["unit"] = unit_name
-            run.iamc.add(df, type=dp_type)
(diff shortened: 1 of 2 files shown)


glatterf42 (approved):
This looks good to me, thanks for the fix :)
How much work do you think it would be to write a test for this function so that we don't accidentally break it in the future?

pmussak:
> This looks good to me, thanks for the fix :) How much work do you think it would be to write a test for this function so that we don't accidentally break it in the future?

@glatterf42  I added 2 smoke tests for the `platform generate` CLI command:
* One to test whether the command runs without errors for an existing platform
* And another whether it errors for a non-existent platform

glatterf42 (approved):
Wonderful, thanks for the test! Now this is even increasing our overall coverage :)

PR #202, opened 2025-09-15
move `configure_logging` call from settings to __main__.py

meksor:
... thus, logging will be configure when using ixmp4 via the cli: `ixmp4 platforms list`, `ixmp4 server start`, but not when used as a library via `import ixmp4`. In the latter case the library user is responsible for setting up the logging configuration...

### DIFF
--- a/ixmp4/conf/settingsmodel.py
+++ b/ixmp4/conf/settingsmodel.py
@@ -57,7 +57,7 @@ class Settings(BaseSettings):
 
         self.setup_directories()
 
-        if _stdout_is_tty or _in_ipython_session:
+        if self.is_in_interactive_mode():
             self.configure_logging(self.mode)
 
         self._credentials: Credentials | None = None
@@ -66,6 +66,9 @@ class Settings(BaseSettings):
         self._manager: ManagerConfig | None = None
 
         logger.debug(f"Settings loaded: {self}")
+
+    def is_in_interactive_mode(self) -> bool:
+        return _stdout_is_tty or _in_ipython_session
 
     @property
     def credentials(self) -> Credentials:
--- a/ixmp4/conf/settingsmodel.py
+++ b/ixmp4/conf/settingsmodel.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import logging.config
+import os
+import sys
 from pathlib import Path
 from typing import Any, Literal
 
@@ -21,6 +23,14 @@ logger = logging.getLogger(__name__)
 
 here = Path(__file__).parent
 root = Path(__root__file__).parent.parent
+
+try:
+    __IPYTHON__  # type: ignore
+    _in_ipython_session = True
+except NameError:
+    _in_ipython_session = False
+
+_stdout_is_tty = os.isatty(sys.stdout.fileno())
 
 
 class Settings(BaseSettings):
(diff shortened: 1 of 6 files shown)


danielhuppmann:
Is there a particular reason against configuring a simple logger with INFO status by default in ixmp4?

meksor:
Yes. 
1. This is unusual for libraries to do.
2. Tomorrow someone might complain that the pyam-compatible version of the logging config is not suitable for their project. Hence go back to 1.

meksor:
Alright ive added these lines:
```py
ixmp4_logger = logging.getLogger("ixmp4")
if ixmp4_logger.level == logging.NOTSET:
    ixmp4_logger.setLevel(logging.INFO)

if len(ixmp4_logger.handlers) == 0:
    ixmp4_logger.addHandler(logging.StreamHandler())
```
I need to explicitly add handlers to the ixmp4 logger so the log messages appear, just setting the level is not enough (maybe thats obvious to logging-aficionados). 
Im sure this will interfere with other stuff as well and if we are doing that we might as well load an entire configuration.  But its either that or configuring manually every time. 

¯\_(ツ)_/¯

meksor:
Another idea that came to my mind was to somehow find out if running in an interactive shell and set up logging if thats the case, some combo of:
```py
try:
    __IPYTHON__
    _in_ipython_session = True
except NameError:
    _in_ipython_session = False
```
and 
```py
import os, sys
if os.isatty(sys.stdout.fileno()):
    ...
```

could work...

glatterf42 (approved):
This looks good to me, though I haven't really been affected. 
FWIW, I'd say we don't need to add (and test and maintain) these lines for setting up logging if running interactively. If people are using ixmp4 via pyam, pyam should employ the level of ixmp4-logging it desires, and if people are using ixmp4 directly, they should take care of that. 
Though if you think it's not too much work to add (and test and maintain) this, users might also appreciate the convenience. No hard opinion from my side here.

ixmp4/conf/settingsmodel.py
```diff
@@ -46,7 +46,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
 
         self.setup_directories()
-        self.configure_logging(self.mode)
+
+        ixmp4_logger = logging.getLogger("ixmp4")
+        if ixmp4_logger.level == logging.NOTSET:
+            ixmp4_logger.setLevel(logging.INFO)
+
+        if len(ixmp4_logger.handlers) == 0:
+            ixmp4_logger.addHandler(logging.StreamHandler())
```
phackstock, line 55: I think in this setup, you'd get log duplication again if the logs are propagated and you log at warning level or above since the root logger is set at WARNING.

danielhuppmann (commented):
So I tested this and I *always* get the usual log output, because `_stdout_is_tty` is always true. I ran the code
```python
import ixmp4
ixmp4.Platform("public-test")
```

in the following cases:
- jupyter notebook
- terminal -> ipython (_in_ipython_session = True)
- terminal -> python (_in_ipython_session = False)
- terminal `python logtest.py` (_in_ipython_session = False)

If this is as you intend it @meksor, fine for me! But I'm wondering which use case would then have no logging with this setup...

meksor:
Youre right, this wasnt quite correct. That version would have turned logging off in a deployment context where the script is not executed from an interactive **system shell** like bash. 
Ive updated the function to check for an interactive **python shell** instead.

danielhuppmann (approved):
So I still get log-output when running test code in the console using `python` and `ipython`, but when running a python script using `python logtest.py`, I only get the unformatted warning, not the info. I guess that this is a good compromise for the library and interactive use cases. Thank you for going down that rabbit hole!

PR #203, opened 2025-09-15
Start transaction id at 2

meksor:
Due to a bug, it is necessary to start the transaction id sequence at 2. 
(the initial version migration inserts a transaction with id=1 even if the database is empty)

### DIFF
--- a/ixmp4/db/migrations/versions/0b45de1602a6_add_version_tables.py
+++ b/ixmp4/db/migrations/versions/0b45de1602a6_add_version_tables.py
@@ -394,7 +394,12 @@ def upgrade():
 
     op.create_table(
         "transaction",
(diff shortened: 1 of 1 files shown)


phackstock (commented):
Looks good to me, just one question. With setting the start to 2, will the initial version migration correctly create a transaction with id=1 or will it just start at 2 and move the issue?

meksor:
It should always create a transaction with id 1.
https://github.com/iiasa/ixmp4/blob/50a887c3255e57218b22e8e289aff9f6ecde5225/ixmp4/db/migrations/versions/84b534bbc858_load_initial_version_data.py#L31

phackstock (approved):
Thanks for the clarification @meksor. Good to be merged from my side then.

danielhuppmann (approved):
Thank you!

PR #204, opened 2025-09-16
add limit(1) to "latest transaction" query

meksor:
Otherwise we always select ALL transactions (20 Million in sc-dev) and throw all but the first away.


danielhuppmann:
Can confirm this works! Thank you!

glatterf42 (approved):
LGTM :)

I'm wondering whether we want to capture this in some kind of test. Less to ensure that this works but more to ensure that this is not accidentally removed in the future at some point.

Also, I'll have to check if I want to apply something similar to the `SELECT`s issued by the optimization rollback.

PR #206, opened 2025-10-08
Adjust to pydantic v2.12

glatterf42:
Today, the CI in the MESSAGE stack broke because of [v2.12 of pydantic](https://github.com/pydantic/pydantic/releases/tag/v2.12.0) (which was released yesterday). 
This release refactors `FieldInfo`, and in https://github.com/pydantic/pydantic/pull/11898#discussion_r2101124381, I found the note that's been causing our errors (as far as I can tell; if we merge this today, tomorrow's CI runs will tell): whenever we do a manual assignment on a `FieldInfo` instance (such as setting `field.json_schema_extra`), we need to also update the dict `field._attributes_set` with the same value.

`poetry update pydantic` also updates some other packages, e.g. mypy, but I hope this just passes without requiring any additional changes.


ixmp4/db/filters.py
```diff
@@ -296,7 +296,9 @@ def filter_func(
                 field = namespace.get(filter_name, Field(None))
             # NOTE field should always be pydantic FieldInfo, convince type checker
             assert field is not None
-            field.json_schema_extra = {"sqla_column": name}
+            json_schema_extra = {"sqla_column": name}
+            field.json_schema_extra = json_schema_extra
+            field._attributes_set["json_schema_extra"] = json_schema_extra
```
meksor, line 301: And this also works with older versions of pydantic? i.e. _attributes_set exists?

meksor (approved):

ixmp4/db/filters.py
glatterf42: Unfortunately, I don't see why the `year__in` filter for datapoints is causing this to fail.

However, I also realized that even a new release of ixmp4 will not fix things for the MESSAGE stack right away because ixmp currently has a pin on the ixmp4 version it can work with. I'll have to address this pin first before a new release here helps anything, which grants us more time to investigate how to properly set up these `alias` values. Though by the above, my hunch is that pydantic might be wrong about saying this parameter doesn't have any effect, in which case we could safely ignore the warnings.

PR #208, opened 2025-10-14
Fixes for ixmp support September/October 2025

glatterf42:
Since you asked about them earlier today, @meksor: these are the changes I have accumulated currently to enable https://github.com/iiasa/ixmp/pull/601. That PR is not ready yet, so it may that further requirements are revealed as I try to enable more ixmp tests to run using ixmp4. However, the further along this process goes, the more detailed these changes are likely to be; so you could already start translating the current state of the `optimization` logic and we'd likely only need minor changes from PRs such as this one.

In other words: maybe I shouldn't have requested a review yet, further changes may come to this PR. Or we merge this as is and I open a new one; either way, reviewing the current set of changes is unlikely to be in vain :)


danielhuppmann (commented):
Just to be sure, please confirm that the new `is_input` column is 
1. not required when adding iamc datapoints using `run.iamc.add()`
2. not included in the dataframe when retrieving data using `run.iamc.tabulate()`

glatterf42:
This is now ready for review :)

PR #232, opened 2026-05-08
Add Migration Version Check for the `DirectTransport` class

### LINKED ISSUE #222: Safeguard against using incompatible DB schemas
As mentioned by @khaeru [here](https://github.com/iiasa/ixmp/pull/611#discussion_r2549017404), the Java backend code safeguards against using incompatible DB schema versions. That is to say, if a user creates and fills a DB using version a.b.c, then updated to version d.e.f, which includes changes to the DB model, the old ixmp code will prevent any errors resulting from that by either 

- raising exceptions right away or 
- performing a migration. 
 
Either option would be fine for us, too, but we should add one of them.

meksor:
When `DirectTransport` (or `AuthorizedTransport`) are instantiated with check_alembic_version=True, it will 
check if the migration version in the database is the same as in the current code base.
For the transport classes the argument is ~`False`~ `True` per default but if instantiated through a platform the value from Settings.check_alembic_version (or the corresponding env var) will be used. Which is also True by default.

Closes #222 

@khaeru Fridolin mentioned this is a request from you so feel free to add you two cents
@danielhuppmann fyi

### Failure Modes

**alembic_version table does not exist:**
```py
raise ImproperlyConfigured(
    "Database schema version check failed because the table "
    "'alembic_version' does not exist. Run migrations or disable the "
    "check via check_alembic_version=False/"
    "IXMP4_CHECK_ALEMBIC_VERSION=false."
)
```
**alembic returns no head revision:**
```py
raise ImproperlyConfigured(
    "Could not determine the expected alembic "
    "revision from migration scripts."
)
```
**alembic returns multiple head revisions**
```py
raise ImproperlyConfigured(
    "Could not determine a unique expected alembic revision because "
    f"multiple heads were found: {head_revision}."
)
```
**alembic_version table has no entries**
```py
raise ImproperlyConfigured(
    "Database schema version check failed because no alembic revision "
    "entry was found in 'alembic_version'. Run migrations or disable "
    "the check via check_alembic_version=False/"
    "IXMP4_CHECK_ALEMBIC_VERSION=false."
)
```
**alembic_version entry is previous revision**
```py
raise ImproperlyConfigured(
    "Database schema version mismatch. "
    f"Expected revision '{head_revision}' but found older "
    f"revision '{current_revision}'. Upgrade the database to the "
    "current ixmp4 migration head, or downgrade ixmp4 to a version "
    "compatible with this database revision."
)
```
**alembic_version entry is unknown revision**
```py
raise ImproperlyConfigured(
    "Database schema version mismatch. "
    f"Expected revision '{head_revision}' but found "
    f"'{current_revision}'. Upgrade your ixmp4 installation or disable the "
    "check via check_alembic_version=False/"
    "IXMP4_CHECK_ALEMBIC_VERSION=false."
)
```

### DIFF
--- a/ixmp4/transport.py
+++ b/ixmp4/transport.py
@@ -5,6 +5,7 @@ import time
 from concurrent.futures import ThreadPoolExecutor
 from email.utils import parsedate_to_datetime
 from functools import lru_cache
+from pathlib import Path
 from typing import Any
 
 import httpx
@@ -15,11 +16,14 @@ from sqlalchemy import orm
 from toolkit.auth.context import AuthorizationContext, PlatformProtocol
 from toolkit.client.auth import Auth, ManagerAuth, SelfSignedAuth
 from toolkit.client.base import ServiceClient
+from toolkit.db.alembic import AlembicController
 
 from ixmp4.base_exceptions import ImproperlyConfigured
 from ixmp4.conf.settings import ClientSettings, Settings
 from ixmp4.core.exceptions import OperationNotSupported, ProgrammingError
 from ixmp4.core.exceptions import registry as exception_registry
+from ixmp4.db import __file__ as db_module_dir
+from ixmp4.db.models import get_metadata
 
 from ._version import __version__
 
@@ -49,8 +53,15 @@ Session = orm.sessionmaker(autocommit=False, autoflush=False)
 class DirectTransport(Transport):
     session: orm.Session
 
-    def __init__(self, session: orm.Session):
+    def __init__(
+        self,
+        session: orm.Session,
+        check_alembic_version: bool = False,
+    ):
         self.session = session
+
+        if check_alembic_version:
+            self.check_alembic_version()
 
         if (url := self.get_database_url()) is not None:
             logger.debug(f"Connected to IXMP4 database at '{url.render_as_string()}'.")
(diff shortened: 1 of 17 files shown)


khaeru:
Thanks for implementing this!

I haven't followed the recent refactor so I do not quite understand the reference to `\*Transport` classes (also don't see them on [this Architecture diagram](https://docs.ece.iiasa.ac.at/projects/ixmp4/en/latest/development/architecture.html#architecture)), but I guess they are intermediaries between a ixmp4…Platform and a ixmp4…Backend: something like ‘DirectTransport’ would do essentially no intermediation, and others might *transport* data between a Platform on a user's machine and a Backend operating on a different machine.

If that's an OK guess, then the PR description makes sense, and it looks like the possible cases of mismatches are all covered, each with a clear message 👍🏾

meksor:
Yes exactly, the refactor introduced a small RPC-like framework which can be used to call procedures of service classes either directly or via the http API on a server. The service classes will internally route calls to their methods depending on which transport class they receive.
The check only runs when creating a direct transport class locally and that is done by the platform automatically if the manager sends you the full DSN (because you are an admin) or if you configure a DSN in the local toml file (via the platforms cli).

ixmp4/transport.py
```diff
@@ -117,6 +129,81 @@ def close(self) -> None:
         assert self.session.bind is not None
         self.session.bind.engine.dispose()
 
+    def get_alembic_controller(self, dsn: str) -> AlembicController:
```
pmussak: `ixmp4.cli.alembic` also defines `get_alembic_controller`, seems equivalent to me. So maybe re-use that or put it in some shared helper.

tests/unit/test_transport.py
pmussak, line 1: As far as I can see none of the tests cover `DirectTransport.from_dsn` with `check_alembic_version=True`.
Is the new error behavior not tested on purpose or just skipped due to the effort involved in adding migration script just for testing or am I missing something else?

pmussak (changes requested):
One DRY suggestion and a question regarding tests

ixmp4/transport.py
meksor: moved it to `ixmp4.db.get_alembic_controller` and moved the env var resolver out everywhere.

tests/unit/test_transport.py
meksor: I originally only tested the constructor directly because from_dsn makes its own session so mocking the appropriate database can get complicated (when in-memory: new connection = new database, and thus no alembic_version table if it was set up naively). 
I added one test that mocks everything in the unit tests, but I figured full integration tests with real temporary sqlite databases are probably also a good idea so I added them as well.

meksor:
Thanks, should all be resolved.

pmussak (approved):
I understand that unit testing transports with `check_alembic_version=True` is a pain, so adding integrations tests sounds like a good idea to me. Lgtm now.

PR #233, opened 2026-05-08
Support filtering by meta indicators 

### LINKED ISSUE #225: Support filtering by meta indicators
Per suggestion by @sorodrigo, it would be useful to add filters to `runs.tabulate()` etc. working on the values of the meta-indicators of a **Run**, eg.
```python
Platform.runs.tabulate(meta={"key": "value"})
```

The current implementation in the scse-components data explorer and the Scenario Compass is to first retrieve the full table of meta-indicators, then downselect to the run-id's where the key-value filters apply, then get the runs for the relevant run-ids- This works but requires more API calls and some user-side processing.

Any thoughts @meksor @pmussak @lenseg?

meksor:
Adds related filters to run and iamc datapoint filters which allow filtering by meta indicators. 
The data layer and api each accept the full meta filter under the "meta" key:

```json
{
  "meta": {
      "key": str,
      "key__in": list[str],
      "key__like": str,
      "key__ilike": str,
      "key__notlike": str,
      "key__notilike": str,
      "dtype": str,
      "dtype__in": list[str],
      // ...  }
}
```

And the facade layer adds shorthand filters: 
```py
mp.runs.tabulate(meta="Emissions Diagnostics|*")
mp.runs.tabulate(meta=[
		"Emissions Diagnostics|Cumulative CCS [2020-2100, Gt CO2]",
		"Climate Assessment|Peak Warming|Median [MAGICCv7.5.3]",
		"Climate Assessment|Year of Peak Warming|Median [MAGICCv7.5.3]"
])
```

Closes #225

@danielhuppmann

### DIFF
--- a/ixmp4/data/filters/__init__.py
+++ b/ixmp4/data/filters/__init__.py
@@ -48,7 +48,34 @@ class DtypeFilter(TypedDict, total=False):
     dtype__in: list[str]
 
 
-class RunMetaEntryFilter(IdFilter, KeyFilter, RunIdFilter, DtypeFilter, total=False):
+class ValueFilter(TypedDict, total=False):
+    value_float: float
+    value_float__lte: float
+    value_float__lt: float
+    value_float__gte: float
+    value_float__gt: float
+    value_float__in: list[float]
+
+    value_int: int
+    value_int__lte: int
+    value_int__lt: int
+    value_int__gte: int
+    value_int__gt: int
+    value_int__in: list[int]
+
+    value_str: str
+    value_str__in: list[str]
+    value_str__like: str
+    value_str__ilike: str
+    value_str__notlike: str
+    value_str__notilike: str
+
+    value_bool: bool
+
+
+class RunMetaEntryFilter(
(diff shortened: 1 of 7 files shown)


danielhuppmann:
This filter implicitly works only on the keys, right? I.e. where a row with a key of that name exists, irrespective of the value.

meksor:
Yes originally the filter only worked for the key and dtype properties. The shorthand/facade filter on the other hand ONLY works on the key field.
I just added more filters for the different value columns:

```py

class ValueFilter(TypedDict, total=False):
    value_float: float
    value_float__lte: float
    value_float__lt: float
    value_float__gte: float
    value_float__gt: float
    value_float__in: list[float]

    value_int: int
    value_int__lte: int
    value_int__lt: int
    value_int__gte: int
    value_int__gt: int
    value_int__in: list[int]

    value_str: str
    value_str__in: list[str]
    value_str__like: str
    value_str__ilike: str
    value_str__notlike: str
    value_str__notilike: str

    value_bool: bool
```

PR #235, opened 2026-05-11
Fix tabulate_manager_platforms printing platform names instead of slugs

meksor:
For the **pyam** workflow we need `tabulate_manager_platforms` to print platform slugs not names. 
Fixes a bug introduced by switching to scse-toolkit (which changes the name and slug attributes).


pmussak (approved):
lgtm

danielhuppmann (approved):
Thank you!

PR #236, opened 2026-05-12
Best Effort Backwards-Compatibilty for <=0.14.0 clients

meksor:
... to fix old ixmp4 versions connecting to new servers, which currently results in a Forbidden exception. 
This PR adds a bunch of compatibilty code to the compat controllers to emulate previous behaviour.

### DIFF
--- a/ixmp4/data/compat_controller.py
+++ b/ixmp4/data/compat_controller.py
@@ -1,11 +1,54 @@
 import json
-from typing import Any, cast
+from typing import Any, Union, cast
 
 from litestar import Request, patch
 
 from ixmp4.core.exceptions import BadRequest
 from ixmp4.data.pagination import GenericPaginatedResult
 from ixmp4.data.services.controller import ServiceController
+
+ExpandType = Union[str, list[str]]
+
+# Filter keys whose values may arrive as plain strings or lists from legacy
+# clients but must be forwarded as nested ``name``-keyed filter dicts.
+NAMED_ENTITY_FILTER_KEYS: frozenset[str] = frozenset(
+    {"region", "variable", "unit", "model", "scenario"}
+)
+
+
+def expand_simple_filter(
+    value: "ExpandType | dict[str, ExpandType]",
+) -> "dict[str, ExpandType]":
+    """Expand a plain string or list filter value into a ``name``-keyed dict.
+
+    Legacy clients send entity filters as bare strings or lists (e.g.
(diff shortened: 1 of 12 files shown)


ixmp4/data/compat_controller.py
```diff
@@ -1,23 +1,72 @@
...
+    {"region", "variable", "unit", "model", "scenario"}
+)
+
+
+def expand_simple_filter(
+    value: "ExpandType | dict[str, ExpandType]",
+) -> "dict[str, ExpandType]":
+    """Expand a plain string or list filter value into a ``name``-keyed dict.
+
+    Legacy clients send entity filters as bare strings or lists (e.g.
```
danielhuppmann, line 24: I'd say that this is more of a general usability feature to have short-hand API, not just a legacy issue. Or would you say that this is something to be moved to pyam and other clients, and remove this compatibility feature at some point?

ixmp4/data/meta/compat_controller.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class RunMetaEntryCompatibilityController(EnumerationCompatibilityController):
+    path = "/"
+
+    def _get_compat_payload(
+        self, query_params: dict[str, Any], body: bytes
+    ) -> dict[str, Any]:
+        payload = super()._get_compat_payload(query_params, body)
+        if "run_id" in payload:
```
danielhuppmann, line 13: Is this something to be fixed in pyam explicitly? Or is this just a ixmp4-backwards-compat issue?

danielhuppmann (approved):
To the extent that I understand this PR, it looks great!

ixmp4/data/compat_controller.py
meksor: I moved the shorthand filters to a set of more explicit `FooFacadeFilter`s since the shorthand filters were undocumented in the data layer and http API (and would result in a 500 response in certain conditions like when a list of wildcard filters is supplied). 
Those filters can be found here: https://github.com/search?q=repo%3Aiiasa%2Fixmp4+symbol%3A%2F%5EFacade.*Filter%2F&type=code
... and currently support the run filter `model` and `scenario` shorthands.

ixmp4/data/meta/compat_controller.py
meksor: just a ixmp4-backwards-compat issue

PR #237, opened 2026-05-12
Fall back to HTTP connections if DB connection is theorethically possible (staff, superusers) but fails anyway

meksor:
Previous ixmp4 versions would never try to connect to the database directly so this was never a problem, now a fallback is needed here. Also refactors all the resolve_dsn_env_tokens calls to where they are immediately needed so leaking dsns with secrets is less likely. DirectTransport now sends a simple ping query upon instantiation to catch any connection errors (can be disabled with ping_database=False).

### DIFF
--- a/ixmp4/conf/platforms.py
+++ b/ixmp4/conf/platforms.py
@@ -108,17 +108,11 @@ class TomlPlatforms(PlatformConnections):
         toml.dump(obj, f)
 
     def list_platforms(self) -> list[TomlPlatform]:
-        return [
-            platform.model_copy(update={"dsn": resolve_dsn_env_tokens(platform.dsn)})
-            for platform in self.platforms.values()
-        ]
+        return list(self.platforms.values())
 
     def get_platform(self, name: str) -> TomlPlatform:
         try:
-            platform = self.platforms[name]
-            return platform.model_copy(
-                update={"dsn": resolve_dsn_env_tokens(platform.dsn)}
-            )
+            return self.platforms[name]
         except KeyError as e:
             raise PlatformNotFound(f"Platform '{name}' was not found.") from e
 
@@ -146,16 +140,10 @@ class ManagerPlatforms(PlatformConnections):
         self.manager_client = manager_client
 
     def list_platforms(self) -> list[Ixmp4Instance]:
-        return [
-            platform.model_copy(update={"dsn": resolve_dsn_env_tokens(platform.dsn)})
-            for platform in self.manager_client.ixmp4.cached_list()
-        ]
+        return self.manager_client.ixmp4.cached_list()
 
     def get_platform(self, name: str) -> Ixmp4Instance:
         for platform in self.manager_client.ixmp4.cached_list():
             if platform.slug == name:
-                return platform.model_copy(
-                    update={"dsn": resolve_dsn_env_tokens(platform.dsn)}
-                )
-        else:
-            raise PlatformNotFound(f"Platform '{name}' was not found.")
+                return platform
+        raise PlatformNotFound(f"Platform '{name}' was not found.")
--- a/ixmp4/conf/platforms.py
+++ b/ixmp4/conf/platforms.py
@@ -36,7 +36,7 @@ def resolve_dsn_env_tokens(dsn: str) -> str:
 
     resolved = _ENV_TOKEN_PATTERN.sub(replace, dsn)
     if missing:
-        logger.error("Missing DSN environment variable(s): " + str(missing))
+        logger.debug("Missing DSN environment variable(s): " + str(missing))
         raise ImproperlyConfigured(
             "Cannot resolve DSN environment variable placeholder(s)."
         )
--- a/ixmp4/core/platform.py
+++ b/ixmp4/core/platform.py
@@ -1,6 +1,13 @@
 import logging
 
-from ixmp4.base_exceptions import PlatformNotFound, PlatformNotUnique
+import sqlalchemy as sa
+
+from ixmp4.base_exceptions import (
+    Ixmp4Error,
+    PlatformNotFound,
+    PlatformNotUnique,
+    ServiceException,
+)
 from ixmp4.conf.platforms import PlatformConnectionInfo
 from ixmp4.conf.settings import Settings
 from ixmp4.data.backend import Backend
@@ -186,7 +193,12 @@ class Platform(object):
             # Try direct connection first
             try:
                 return DirectTransport.from_dsn(ci.dsn)
-            except Exception as e:
+            except (
+                ServiceException,
+                Ixmp4Error,
+                sa.exc.SQLAlchemyError,
+                ImportError,
+            ) as e:
                 # If direct connection fails and HTTP URL is available,
                 # fall back to HTTP transport.
                 if ci.url is not None:
--- a/ixmp4/core/platform.py
+++ b/ixmp4/core/platform.py
@@ -202,8 +202,11 @@ class Platform(object):
                 # If direct connection fails and HTTP URL is available,
                 # fall back to HTTP transport.
                 if ci.url is not None:
+                    logger.debug(
+                        f"Error while trying to establish direct connection: \n{e}"
+                    )
                     logger.warning(
-                        f"Direct connection failed: {e}. "
+                        f"Direct connection failed with `{e.__class__.__name__}`. "
                         "Falling back to HTTP connection for platform "
                         f"'{ci.name}' at {ci.url}"
                     )
--- a/ixmp4/core/platform.py
+++ b/ixmp4/core/platform.py
@@ -183,7 +183,27 @@ class Platform(object):
                 auth=self.settings.get_client_auth(cred_dict),
             )
         else:
-            return DirectTransport.from_dsn(ci.dsn)
+            # Try direct connection first
+            try:
+                return DirectTransport.from_dsn(ci.dsn)
+            except Exception as e:
+                # If direct connection fails and HTTP URL is available,
+                # fall back to HTTP transport.
+                if ci.url is not None:
+                    logger.warning(
+                        f"Direct connection failed: {e}. "
+                        "Falling back to HTTP connection for platform "
+                        f"'{ci.name}' at {ci.url}"
+                    )
+                    cred_dict = self.settings.get_credentials().get(http_credentials)
+                    return HttpxTransport.from_url(
+                        str(ci.url),
+                        settings=self.settings.client,
+                        auth=self.settings.get_client_auth(cred_dict),
+                    )
+                else:
+                    # No HTTP URL available, re-raise the original error
+                    raise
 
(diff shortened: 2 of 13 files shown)


ixmp4/conf/platforms.py
```diff
@@ -146,16 +140,10 @@ def __init__(self, manager_client: ManagerClient):
...
 
     def get_platform(self, name: str) -> Ixmp4Instance:
         for platform in self.manager_client.ixmp4.cached_list():
             if platform.slug == name:
-                return platform.model_copy(
-                    update={"dsn": resolve_dsn_env_tokens(platform.dsn)}
-                )
-        else:
-            raise PlatformNotFound(f"Platform '{name}' was not found.")
+                return platform
```
danielhuppmann, line 148: I noticed that this iterates over the names in the list. Is there a guard that no two platforms with the same name can exist?

danielhuppmann (approved):
Looks good to the extent that my technical capacity matches the code complexity

ixmp4/conf/platforms.py
meksor: Yes, the manager has a database unique constraint.

danielhuppmann:
Tried this in production now, and the log output when connecting via API looks like

```
[INFO] 12:19:32 - ixmp4.core.platform: PRISMA Internal: This is the project-internal platform for the PRISMA project.
[ERROR] 12:19:32 - ixmp4.conf.platforms: Missing DSN environment variable(s): {'IXMP4_POSTGRES_PASSWORD'}
[WARNING] 12:19:32 - ixmp4.core.platform: Direct connection failed: Cannot resolve DSN environment variable placeholder(s).. Falling back to HTTP connection for platform 'PRISMA Internal' at https://ixmp4.ece.iiasa.ac.at/v1/prisma-internal/
[WARNING] 12:19:34 - ixmp4.transport: IXMP4 Client and Server versions do not match. (Client: 0.0.0, Server: 0.15.6)
[INFO] 12:19:34 - ixmp4.data.backend: Creating backend class with transport: <HttpxTransport base_url=https://ixmp4.ece.iiasa.ac.at/v1/prisma-internal/ user=<User id=104 username=huppmann>>
```

Seems a bit excessive to show an error when this is expected behaviour?

And from a standard-user perspective, can we streamline the log to something like
> Connecting to ixmp4 Platform `prisma-internal` as user `huppmann`

meksor:
That's true. I'll change the logger call in resolve_dsn_env_tokens to a debug level.
Keep in mind that non staff, non superusers will not see the failed DB Connection since they will attempt http directly.

ixmp4/core/platform.py
```diff
@@ -183,7 +183,27 @@ def get_transport(
                 auth=self.settings.get_client_auth(cred_dict),
             )
         else:
-            return DirectTransport.from_dsn(ci.dsn)
+            # Try direct connection first
+            try:
+                return DirectTransport.from_dsn(ci.dsn)
+            except Exception as e:
```
pmussak: If the direct connection fails and the HTTP transport succeeds, this will hide underlying errors like `ProgrammingError` which is a subclass of `Exception`.
Maybe narrow it to whatever sqlalchemy error we expect and/ or `ImproperlyConfigured`, since the intent is to catch improperly configured ixmp4 clients, right (namely once that do not have the DB password set as ENV var)?

ixmp4/core/platform.py
```diff
@@ -183,7 +183,27 @@ def get_transport(
...
+                # If direct connection fails and HTTP URL is available,
+                # fall back to HTTP transport.
+                if ci.url is not None:
+                    logger.warning(
+                        f"Direct connection failed: {e}. "
```
pmussak: At this point the dsn already includes secrets, (interpolated in the `DirectTransport.from_dsn(ci.dsn)` call).
I don't know, but if the underlying sqlalchemy exceptions includes the dsn in the error message, this will leak it into the log.

ixmp4/transport.py
```diff
@@ -133,10 +139,9 @@ def __init__(
...
         platform: PlatformProtocol,
+        ping_database: bool = True,
     ):
-        super().__init__(
-            session,
-        )
+        super().__init__(session, ping_database=ping_database)
         self.auth_ctx = auth_ctx
         self.platform = platform
         self.unauthorized_transport = DirectTransport(session)
```
pmussak: I think this will ping the DB twice on every `AuthorizedTransport` init.
Once via `super().__init__(session, ping_database=ping_database)` and a second time via `DirectTransport(session)`.
This seems unnecessary to me, but maybe I am missing something.

pmussak: Also why not pass
```suggestion
        self.unauthorized_transport = DirectTransport(session, ping_database)
```

ixmp4/transport.py
```diff
@@ -38,6 +39,7 @@ def __str__(self) -> str:
 @lru_cache()
 def cached_create_engine(dsn: str, **kwargs: Any) -> sa.Engine:
     # max_identifier_length=63 to avoid exceeding postgres' default maximum
+    dsn = resolve_dsn_env_tokens(dsn)
     return sa.create_engine(
         dsn, poolclass=sa.NullPool, max_identifier_length=63, **kwargs
     )
```
pmussak, line 44: I see a potentially hard to debug problem here. The cache uses the dsn with unresolved env vars. So if the env var is changed (DB passwd changed) and the cache stays intact * the cache will resolve to an engine using the old dsn. 

*I am not sure whether this condition can ever be met, but maybe you know @meksor

pmussak (changes requested):
I think this is a sound patch-fix implementation for handling the case where staff- or superusers use ixmp4 in client mode and do not have the necessary env vars set.

But this whole `resolve_dsn_env_tokens` business seems dangerous to me. I noted some places where I think this might cause errors or leaks.
Do you have a mid-term goal for handling this more safely?

ixmp4/core/platform.py
meksor: Yeah, I tried that already so I have an approach, but I worry that I will need to constantly update this part of the code so new exceptions that I previously didn't encounter are handled. For example even switching from psycopg (which unintuitively contains versions 1 and 3 as far as I remember) to psycopg2 or asyncpg will introduce different DBAPI exceptions that will not be caught (sqlalchemy does not wrap many of those exceptions). Since this can potentially be done purely by configuring a database DSN and installing the driver alongside IXMP4 I opted to catch all exceptions for this relatively small call (.from_dsn + creating the engines + Transport constructor).

But I will reasses this and it's environment and the feasibility of handling this differently and get back to this shortly.

ixmp4/core/platform.py
meksor: I don't think it does since the sqlalchemy.URL class hides the password unless instructed otherwise and I have seen these errors with hidden passwords, but I will double check.

ixmp4/transport.py
meksor: I guess I would always explicitly pass `ping_database=False` so the ping isn't performed twice if True as you said.

ixmp4/core/platform.py
meksor: > since the intent is to catch improperly configured ixmp4 clients, right (namely once that do not have the DB password set as ENV var)?

Well thats one reason for failure, but the password might also be wrong or the database is unreachable, which are DBAPI exceptions mentioned above.

meksor: Updated the handling logic, all the failures I am able to produce do get wrapped by SQLAlchemyError. Maybe my info about that is outdated...

ixmp4/core/platform.py
meksor: Looks like sqlalchemy logs the exceptions at a DEBUG level as well, so I followed that approach and changed the WARNING level message to only mention the exception class name

ixmp4/transport.py
meksor: I set ping_database=False for the unauthorized_transport.

ixmp4/transport.py
meksor: I think it is not likely right now, but the ixmp4 Settings object potentially calls load_dotenv internally (not sure how pydantic does that exactly) and thus might update os.environ.
In any case thats a good catch because if we transition to another secrets-management technique, we will need to watch this caching bug anyway. 
Fixed.

meksor:
> But this whole resolve_dsn_env_tokens business seems dangerous to me. I noted some places where I think this might cause errors or leaks.
Do you have a mid-term goal for handling this more safely?

Well, I think this OK right now. The way it was done before was quite dangerous because the platform objects with substituted env vars was passed around. 
Two possible improvements in my view:
1. Splitting up the string in the manager data model, so having driver, user, host, db and  "password_envvar" as seperate properties instead of just one string, but then one needs two different data models for postgres and sqlite databases (and one needs to settle on what is substitutable). 
2. Using the new IIASA password management service to obtain the secrets. 
@pmussak

meksor:
The updated logging now looks like this:

@danielhuppmann 
```py
>>> import ixmp4
>>> mp = ixmp4.Platform("dev-public")
[INFO] 16:38:57 - ixmp4.core.platform: IXMP4 Dev Public: Public test platform.
[WARNING] 16:38:57 - ixmp4.core.platform: Direct connection failed with `ImproperlyConfigured`. Falling back to HTTP connection for platform 'IXMP4 Dev Public' at https://dev.ixmp.ece.iiasa.ac.at/v1/dev-public
[WARNING] 16:38:59 - ixmp4.transport: IXMP4 Client and Server versions do not match. (Client: 0.0.0, Server: 0.15.7)
[INFO] 16:38:59 - ixmp4.data.backend: Creating backend class with transport: <HttpxTransport base_url=https://dev.ixmp.ece.iiasa.ac.at/v1/dev-public/ user=<User id=101 username=wolschlager>>
```

Where normal users will not see the first warning line and users with an up to date ixmp4 will not see the second.

ixmp4/core/platform.py
pmussak: Nice to see that you were able to narrow this catch block.

pmussak (approved):
Changes look good to me, thanks for quickly looking into all my comments!

Regarding your reaction to my initial review:
> > But this whole resolve_dsn_env_tokens business seems dangerous to me. I noted some places where I think this might cause errors or leaks.
> > Do you have a mid-term goal for handling this more safely?
> 
> Well, I think this OK right now. The way it was done before was quite dangerous because the platform objects with substituted env vars was passed around. Two possible improvements in my view:

I see, I did not have in mind that before the platform object with substituted env vars was passed around. Then this is definitely a huge improvement.

>     1. Splitting up the string in the manager data model, so having driver, user, host, db and  "password_envvar" as seperate properties instead of just one string, but then one needs two different data models for postgres and sqlite databases (and one needs to settle on what is substitutable).
> 
>     2. Using the new IIASA password management service to obtain the secrets.

My gut feeling is that waiting for a couple of months and then implementing something with the new secret management service is the way to go.
The other option sounds like potential config nightmare.

PR #238, opened 2026-05-14
Fix Exception Logging Behaviour

meksor:
Currently the server is only able to log exceptions in debug mode. 
This adds a configuration variable `ServerSettings.log_exceptions`/`IXMP4_SERVER__LOG_EXCEPTIONS` which controls the litestar logging config. It can be "always" (default) "never" or "debug". [litestar docs](https://docs.litestar.dev/2/reference/logging/config.html#litestar.logging.config.LoggingConfig.log_exceptions)
Some status codes are never logged as they are expected client exceptions: 422, 409, 404, 403, 401, 400


phackstock (approved):
lgtm 👍

PR #240, opened 2026-05-18
Fix Server Exception Handler for toolkit exceptions

meksor:
Fixes bad behaviour for "InvalidToken" exceptions which currently just become non-descript 500 Internal Server Error responses. The basic toolkit exceptions were unhandled, thus the problem. 
Changes the root exception type for the handler and moves the handler to the top level server object (from the v1 api). Adds tests to assert correct behaviour.
Also adds `__init.py__` files to the test modules so pytest does not get confused with multiple test_server modules.


pmussak (approved):
lgtm

PR #243, opened 2026-05-26
Fix accidental masking of DSN passed to AlembicController

meksor:
Fixes a bug I noticed in #242 which causes the alembic version check to fail if the database DSN contains a secret.
Adds `hide_password=False` to the render_as_string call so the full unmasked dsn is used.
Also adds a unit test to assert that it works.


danielhuppmann (approved):
Not sure about the technical details, but can confirm that switching to this branch allowed me to connect to a local database (where the latest release didn't work).

PR #249, opened 2026-06-03
Fix run.optimization.<item>.list() listing items from all runs.

meksor:
Also adds tests that assert the bug stays fixed


danielhuppmann (approved):
Important fix!

PR #251, opened 2026-06-08
Accept mixed time column in IAMC data

meksor:
Proposes a simple adjustment to accept a "time" column in IAMC dataframes which can contain both integer/year and datetime values. If the value is an integer the "subannual" (or category or step_category) column may contain a subannual category string. Adds a test to assert new behavior.

### DIFF
--- a/ixmp4/core/iamc/data.py
+++ b/ixmp4/core/iamc/data.py
@@ -42,9 +42,14 @@ class IamcDataFacade(object):
     def _split_time_col(df: pd.DataFrame) -> pd.DataFrame:
         time = df["time"]
         is_year = time.apply(lambda x: isinstance(x, (int, np.integer)))
-        df["year"] = pd.Series(np.nan, index=df.index, dtype="object")
-        df.loc[is_year, "year"] = time.loc[is_year]
-        df["datetime"] = pd.to_datetime(time.where(~is_year), errors="coerce")
+
+        if is_year.any():
+            df["year"] = pd.Series(np.nan, index=df.index, dtype="object")
+            df.loc[is_year, "year"] = time.loc[is_year]
+
+        if not is_year.all():
+            df["datetime"] = pd.to_datetime(time.where(~is_year), errors="coerce")
+
         return df.drop(columns=["time"])
 
     @classmethod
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Thank you, uses this branch for a much more concise `to_ixmp4()` implementation in pyam and it worked like a charm!

ixmp4/core/iamc/data.py
```diff
@@ -40,11 +40,11 @@ def _rename_arg_cols(df: pd.DataFrame) -> pd.DataFrame:
...
     def _split_time_col(df: pd.DataFrame) -> pd.DataFrame:
-        is_year = df["time"].apply(lambda x: isinstance(x, (int, np.integer)))
-        df["year"] = np.where(is_year, df["time"], np.nan)
-        df["datetime"] = pd.to_datetime(
-            np.where(~is_year, df["time"], "NaT"), errors="coerce"
-        )
+        time = df["time"]
+        is_year = time.apply(lambda x: isinstance(x, (int, np.integer)))
+        df["year"] = pd.Series(np.nan, index=df.index, dtype="object")
+        df.loc[is_year, "year"] = time.loc[is_year]
```
danielhuppmann: I guess you could add a little shortcut?
```suggestion
    if any(is_year):
           df["year"] = pd.Series(np.nan, index=df.index, dtype="object")
           df.loc[is_year, "year"] = time.loc[is_year]
```

meksor: Good idea, I used the numpy `series.all/any` for better performance and added an inverse condition for the datetime column:
```py
    @staticmethod
    def _split_time_col(df: pd.DataFrame) -> pd.DataFrame:
        time = df["time"]
        is_year = time.apply(lambda x: isinstance(x, (int, np.integer)))

        if is_year.any():
            df["year"] = pd.Series(np.nan, index=df.index, dtype="object")
            df.loc[is_year, "year"] = time.loc[is_year]

        if not is_year.all():
            df["datetime"] = pd.to_datetime(time.where(~is_year), errors="coerce")

        return df.drop(columns=["time"])

```

PR #252, opened 2026-06-08
Harmonize all string-representations of objects

danielhuppmann:
I thought that having the id first and without clear identification of its meaning wasn't really helpful to average users like me... Tried to do this consistently through all similar items.


meksor (approved):
Looks good!

PR #257, opened 2026-06-25
Quickfix for meta in datetime format

### LINKED ISSUE #256: Improve error message when trying to add non-supported data type as meta data entry
Currently we're getting some sql error when trying to submit a non-supported data type (such as datetime) as a meta indicator value.
It would be good to either:
* Raise a better error message _or_
* Cast everything that's non-supported to string and save it that way

Of course the casting to string could also fail so maybe we'll have to do both anyway.

danielhuppmann:
This PR fixes a database-error when trying to write a meta-indicator in a unsupported format such as datetime (currently supported int, float, str, bool).

The solution casts any non-supported type to string and saves it in this format - not reverting to the original state when reading from the database. We will treat that as a known limitation for the time being.

Tests currently fail because the RestAPI-route casts the pd.Timestamp to a string "2026-06-25T00:00" whereas the direct route casts to "2026-06-25 00:00" (whitespace instead of "T").

Closes #256

### DIFF
--- a/ixmp4/core/meta.py
+++ b/ixmp4/core/meta.py
@@ -14,8 +14,8 @@ from ixmp4.data.meta.filter import (
 )
 from ixmp4.data.meta.service import RunMetaEntryService
 
-from .base import BaseServiceFacade
 from ..data.meta.type import convert_value
+from .base import BaseServiceFacade
 
 if TYPE_CHECKING:
     from ixmp4.data.backend import Backend
--- a/ixmp4/core/meta.py
+++ b/ixmp4/core/meta.py
@@ -116,8 +116,7 @@ class RunMetaDictFacade(
 
         py_value = numpy_to_pytype(value)
         if py_value is not None:
-            _, py_value = convert_value(py_value)
-            self._service.create(self.run.id, key, py_value)
+            self._service.create(self.run.id, key, convert_value(py_value))
         self.df, self.data = self._get()
 
     def __delitem__(self, key: str) -> None:
@@ -200,7 +199,7 @@ class RunMetaDescriptor(object):
             {"key": value.keys(), "value": [numpy_to_pytype(v) for v in value.values()]}
         )
         df.dropna(axis=0, inplace=True)
-        df["value"] = df["value"].map(lambda v: convert_value(v)[1])
+        df["value"] = df["value"].map(lambda v: convert_value(v))
         df["run__id"] = obj._dto.id
         obj._backend.meta.bulk_upsert(df)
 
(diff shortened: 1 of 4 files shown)


danielhuppmann:
FYI @phackstock

danielhuppmann:
@meksor, any thoughts on how to reconcile this? Tried to test only with a date (to avoid the T) but the httpx-casting is too smart for that. And I didn't find where the api-conversion is happening, I assume that this is now all abstracted away?

meksor:
Well, pd.Timestamp is another complicated type (different than datetime) and the serialization within dataframes (which is the path taken here) is handled by `ixmp4/data/dataframe.py`. I tried to handle this in your branch in a few different ways but I think the best way to do it is to add a `convert_value` function to the meta/type.py module:

```py
def convert_value(value: Any) -> tuple[Type, Any]:
    value_type = type(value)

    if value_type in _type_map:
        return _type_map[value_type], value

    logger.debug(
        "Converting value of type '%s' to string.", value_type.__name__
    )
    try:
        return Type.STR, str(value)
    except Exception as e:
        raise InvalidRunMeta(
            f"Failed to convert value of type '{value_type.__name__}' to string: {e}"
        ) from e
```
And then use it in the facade layer 

```py
class RunMetaDictFacade: ...
    def __setitem__(self, key: str, value: MetaValueType | np.generic | None) -> None:
        self.run.require_lock()

        try:
            del self[key]
        except KeyError:
            pass

        py_value = numpy_to_pytype(value)
        if py_value is not None:
            _, py_value = convert_value(py_value)
            self._service.create(self.run.id, key, py_value)
        self.df, self.data = self._get()

class RunMetaDescriptor: ...
    def __set__(
        self, obj: "Run", value: Mapping[str, MetaValueType | np.generic | None]
    ) -> None:

        obj.require_lock()
        self._delete_existing(obj)

        df = pd.DataFrame(
            {"key": value.keys(), "value": [numpy_to_pytype(v) for v in value.values()]}
        )
        df.dropna(axis=0, inplace=True)
        df["value"] = df["value"].map(lambda v: convert_value(v)[1])
        df["run__id"] = obj._dto.id
        obj._backend.meta.bulk_upsert(df)

```

And throw `InvalidRunMeta` in the metas `ItemRepository` and `PandasRepository` or in the `Type` class. If the type is invalid so even old clients get an OK error message. 
 

PS: I guess returning the Type from convert_value is unecessary.

danielhuppmann:
Thank you @meksor for the assist, learned quite a bit today!

But in the process, this became a bit more complex than I think is reasonable, so I implemented the alternative suggestion by @phackstock, just raising a clear error, see #258. 

Now that I see both, I like #258 better (because it keeps the codebase cleaner), but I let @meksor and @phackstock make the call which (or if any) of the PRs to accept.

phackstock:
Thanks for the work @danielhuppmann and @meksor, my preference is for keeping things simple and having the user submit data in the proper format so my preference is for #258

danielhuppmann:
Closing in favour of #258

PR #258, opened 2026-06-25
Add simple check for meta-indicator value type

### LINKED ISSUE #256: Improve error message when trying to add non-supported data type as meta data entry

danielhuppmann:
This is an alternative to #257, raising a clear error message when meta-indicator values of an unsupported type are added instead of (trying to) casting everything unexpected to string.

Closes #256

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     entry: bash -c "poetry run mypy ."
     language: system
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.0
+  rev: v0.15.20
   hooks:
   - id: ruff
   - id: ruff-format
--- a/ixmp4/core/meta.py
+++ b/ixmp4/core/meta.py
@@ -115,8 +115,9 @@ class RunMetaDictFacade(
             pass
 
         py_value = numpy_to_pytype(value)
+        check_meta_type(py_value)
         if py_value is not None:
-            self._service.create(self.run.id, key, check_meta_type(py_value))
+            self._service.create(self.run.id, key, py_value)
         self.df, self.data = self._get()
 
     def __delitem__(self, key: str) -> None:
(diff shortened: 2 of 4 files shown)


danielhuppmann:
I give up, I don't know which ruff version to use to make the linter pass...

meksor:
The version should not matter or the version in the poetry.lock file should work. 
At the bottom of the CI job it says:

Would reformat: tests/core/test_meta.py
1 file would be reformatted, 319 files already formatted

Did you try: ruff format tests/core/test_meta.py ?

Edit: My IDE also sorts imports for me which you may be able to achieve using :
ruff check --select I --fix

danielhuppmann:
> Did you try: ruff format tests/core/test_meta.py ?

Of course, and I tried it with both 0.15.12 (per `poetry.lock`) and 0.15.20 (latest) - locally, all my files are fine...

meksor:
My IDE also sorts imports for me which you may be able to achieve using :
ruff check --select I --fix

danielhuppmann:
image

meksor:
I cleared the cached virtualenv using the github cli and restarted the Code Quality job. 
Sometimes this cache bugs out and generates lots of errors like before, not sure why. 
Anyway, now we can also see two mypy errors 

```
tests/core/test_meta.py:55: error: Dict entry 1 has incompatible type "str": "Timestamp"; expected "str": "float | int | str | generic[Any] | None"  [dict-item]
tests/core/test_meta.py:59: error: Incompatible types in assignment (expression has type "Timestamp", target has type "bool | float | int | str | generic[Any] | None")  [assignment]
```

meksor:
Wow, thats annoying... 
I am not sure how exactly this pre-commit thing works since fridolin set it up, but if it really is a version mismatch (pre-commit is a few versions behind) then you can try changing `.pre-commit-config.yaml` line `9`  to   `rev: v0.15.12`. Maybe I was too confident in ruff backwards compatibility.

meksor:
#259

meksor (approved):
lgtm

phackstock (approved):
Good to be merged from my side.

PR #259, opened 2026-06-26
Update mypy and ruff versions for pre-commit

meksor:


danielhuppmann:
I'm wondering if a simple ruff action wouldn't be a lot easier...?
https://github.com/IAMconsortium/pyam/blob/main/.github/workflows/ruff.yml

meksor:
Thats true! #260 is an alternative.

Issue #215, opened 2025-11-06
`inf` value handling in the iamc api


meksor:
Currently, ixmp4 allows inf values to be added for IAMC datapoints. There seem to be problems with serializing these values to json via the web api and filtering by them.

1. Should inf values even be allowed? 
2. If yes, how do we represent them in json and efficiently serialize? 
3. How do we filter by them, how are they used in the databases?

meksor:
Great, that makes it easy

PR #266, opened 2026-09-01
Cross-platform `Run` cloning

meksor:
As per an internal discussion, this PR implements **cross-platform** cloning. As in cloning a run from Platform A to Platform B. No restrictions are made toward which platforms serve as source and destination. Local, HTTP and centralized PostgreSQL platforms can all be copied to and from.

Run.clone() now accepts `platform : Platform | Backend | None`. When given, the destination run is created on that platform's backend with `backend.runs.create(...)` instead of `self._service`/`self._backend`. When omitted, the clone is created on the source run's platform as before. No code changes are needed beside that one. 

`tests/core/test_run.py` is also expanded by a test that creates an independent platform and clones a run to it. 

While writing the docstring for this feature, I noticed that the docstring for `TimeSeriesService.bulk_upsert` was inaccurate, so I fixed it.


Wegatriespython:
Thanks for the PR @meksor. One concern I found is around failure handling when cloning across platforms.

For MESSAGE workflows, cross-platform cloning is fairly error-prone because units and regions need to exist on the destination first. Ideally the clone should therefore be atomic: either it succeeds completely or leaves nothing behind.

I tested this with the R12 MESSAGEix-GLOBIOM SSP2 v6.1 baseline: 13.2 million parameter rows across 68 parameters. I loaded it into a local platform and cloned it to a second platform created with `ixmp4 platforms add`.

The copy itself looks solid and performs about the same as a same-platform clone: 92s to another SQLite platform vs. 97s for `run.clone(scenario=...)` on the same platform.

Cloning to a freshly created platform fails after ~15s with `UnitNotFound: '' is not defined for this Platform!`, but leaves behind a run containing 25 index sets and one parameter. Retrying creates another partial run. After copying the units across, the clone succeeds, leaving three runs on the destination, two of which need to be deleted manually.

Since a fresh platform has no units, this is likely to be a common first-time failure rather than an edge case. Checking required units and regions before creating the destination run would avoid the residue and could report all missing items at once.

I also hit a separate PostgreSQL issue. Cloning the same run fails with `DatabaseError: Execution failed on sql 'SELECT checkpoint.run__id checkpoint.message, checkpoint.id ...`. Afterwards the platform is unusable for the rest of the process with `FailedSqlTransaction: current transaction is aborted`.

The underlying issue does not seem specific to cloning. This works on SQLite but fails on PostgreSQL:

```python
df = pd.DataFrame({"i": ["a", "b", "c"], "values": [1.0, float("-inf"), float("inf")], "units": ["", "", ""]})
par.add_data(df)
```

Real MESSAGE scenarios contain these values. The R12 baseline has `bof` in 13,757 rows, so it can currently be stored on SQLite but not PostgreSQL.

Happy to open the PostgreSQL/`inf` behaviour as a separate issue if it is not already known.

meksor:
Hi, thanks for your comments. I have added an automatic "dirty run cleanup", one caveat: deleting runs requires a management permission level while creating it requires only edit permissions. That means a dirty run cannot be deleted when the user only has edit permission, instead a warning is emitted.

meksor:
There is a discussion about `inf` values on the IAMC side [here](https://github.com/iiasa/ixmp4/issues/215). 
Since this is a different use-case though we probably need to find some workaround for [PostgreSQL/JSONs no-inf limitation](https://www.postgresql.org/docs/current/datatype-json.html). 
For clarification: Does the sign (`+inf` vs `-inf`) need to be preserved? Are you storing `None`/`NaN` values as well?

danielhuppmann:
FWIW, in pyam's `write_to_ixmp4()`, there is a check [here](https://github.com/IAMconsortium/pyam/blob/6a1b2a2d799971bc67a8d3252b40d5640bb23c18/pyam/ixmp4.py#L177) that all regions and units exist in the platform before creating a new **Run** - might be worthwhile here rather than cleaning up later.

danielhuppmann:
Let's please have the `inf` discussion separately - I thought that I had introduced flag-parameters in the GAMS code to avoid inf...?

meksor:
I added a pessimistic check when cloning to other platforms. Might make the performance slightly worse since all the iamc data, scalars and parameters are checked; adding some overhead.

Wegatriespython:
Ah great, I will rerun the clone benchmark to measure the consequence

Wegatriespython:
Okay I ran the check with a R12 scenario clone. The check costs around 20s. There is this issue which I discovered: the old error message mentioned the missing unit. `UnitNotFound: '' is not defined for this Platform!` but the new tells `UnitNotFound: Unit not found`.  It seems the empty string is causing default raise fallback, would be better propagate the empty string (which is a real value on message-ix scenarios)

PR #268, opened 2026-09-16
Fix ValueError with scalar Variable

### LINKED ISSUE #267: ValueError: No group keys passed! with scalar Variable
In iiasa/ixmp#640, a call to `.core.optimization.variable.Variable.add_data()` produces ValueError in `.data.optimization.base.repositories.IndexedRepository.validate_data()`:
```python
ixmp/backend/ixmp4_io.py:376: in _read_variables_to_run
    variable.add_data(records[columns_of_interest])
../../../.venv/3.14/lib/python3.14/site-packages/ixmp4/core/optimization/variable.py:92: in add_data
    self._service.add_data(id=self._dto.id, data=data)
../../../.venv/3.14/lib/python3.14/site-packages/ixmp4/data/services/procedure/__init__.py:197: in wrapper
    return cast(ReturnT, auth_callable(*direct_args, **direct_kwargs))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.venv/3.14/lib/python3.14/site-packages/ixmp4/data/optimization/variable/service.py:279: in add_data
    self.items.add_data(id, data)
../../../.venv/3.14/lib/python3.14/site-packages/ixmp4/data/optimization/base/repositories.py:51: in add_data
    self.validate_data(
../../../.venv/3.14/lib/python3.14/site-packages/ixmp4/data/optimization/base/repositories.py:202: in validate_data
    if data[limited_to_indexsets.keys()].value_counts().max() > 1:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.venv/3.14/lib/python3.14/site-packages/pandas/core/frame.py:8691: in value_counts
    counts = self.groupby(
../../../.venv/3.14/lib/python3.14/site-packages/pandas/util/_decorators.py:336: in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
../../../.venv/3.14/lib/python3.14/site-packages/pandas/core/frame.py:10835: in groupby
    return DataFrameGroupBy(
../../../.venv/3.14/lib/python3.14/site-packages/pandas/core/groupby/groupby.py:1095: in __init__
    grouper, exclusions, obj = get_grouper(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = Empty DataFrame
Columns: []
Index: [0], key = [], level = None, sort = False, observed = False, validate = True, dropna = True

    def get_grouper(
        obj: NDFrameT,
        key=None,
        level=None,
        sort: bool = True,
        observed: bool = False,
        validate: bool = True,
        dropna: bool = True,
    ) -> tuple[ops.BaseGrouper, frozenset[Hashable], NDFrameT]:

        # ...
    
        if len(groupings) == 0 and len(obj):
>           raise ValueError("No group keys passed!")
E           ValueError: No group keys passed!

../../../.venv/3.14/lib/python3.14/site-packages/pandas/core/groupby/grouper.py:929: ValueError
```

In particular, this occurs with:
- A 'scalar' or 0-dimensional Variable.
- Input data (`records[columns_of_interest]` in the snippet) like:
  ```
     levels  marginals
  0     1.0        0.0
  ```
- ixmp4 v0.16.8
- Python 3.14

khaeru:
Closes #267.


khaeru:
@meksor I've looked at [DEVELOPING.md](https://github.com/iiasa/ixmp4/blob/main/DEVELOPING.md#contributing) and have run pre-commit on the changes whilst committing.

I don't see any specific other instructions to follow, but if I have missed something, please tell me what.

For instance, I thought about adding tests for changed behaviours or fixed bugs, which is a common request for other projects. I read the [.test module docstring](https://github.com/iiasa/ixmp4/blob/main/tests/__init__.py), and figured out that perhaps `.tests.data.test_optimization_variable_service` might be the right place for a test of adding data to a 0-D variable, but could not figure out how or where to add it. If that is something you'd like to have, please feel free to co-assign yourself on the PR, or give instructions. In lieu, I ran the test suite for iiasa/ixmp#640 with these changes and can confirm that the error reported in #267 does not occur.

meksor:
Looks good, thanks for the fix. 
Test: yes please, there are two relevant test files:

data/test_optimization_variable.py 
And 
core/test_optimization_variable.py

Testing the data and facade layers respectively. Looks like core/ already has a 0D test case, but no data is added which is why it doesn't fail (presumably).

The data layer test can be a new test class (inheriting the base in that file, like the others). The core test could be implemented by just adding data to the variables in the first test class which already creates a 0D var.

I need at least one of these to merge these. Which one or both is your decision!

khaeru:
> there are two relevant test files:
> 
> data/test_optimization_variable.py
> And
> core/test_optimization_variable.py
>
>
> Testing the data and facade layers respectively.

I guess the first file should be data/test_optimization_variable_service.py.

> Looks like core/ already has a 0D test case

I believe you, but finding a "0D test case" in those ~2000 lines, without comments, defeated me. I'll try again.

meksor:
> I guess the first file should be data/test_optimization_variable_service.py.

Yes

> I believe you, but finding a "0D test case" in those ~2000 lines, without comments, defeated me. I'll try again.

A variable with no indexsets/columns is created in TestVariable.test_create_variable on line 43. But now I am looking at the file again it might make more sense to add another subclass of VariableDataTest which overrides the test_data_indexsets fixture in addition to implementing the other ones. 
This is a common pattern in the tests to write a base class that defines tests and parameterize them by implementing subclasses that implement the test data fixtures. 

Was able to quickly reproduce the original ValueError on a different branch with this test, but didnt test fully if deleting by levels/marginals works:

```py
class TestZeroDimensionVariableData(VariableDataTest):
    @pytest.fixture(scope="class")
    def test_data_indexsets(self) -> list[ixmp4.optimization.IndexSet]:
        return []

    @pytest.fixture(scope="class")
    def column_names(
        self,
    ) -> list[str] | None:
        return None

    @pytest.fixture(scope="class")
    def test_data(self) -> dict[str, list[Any]] | pd.DataFrame:
        return {
            "marginals": [-2, 1, 1],
            "levels": [2, 1, 3],
        }

    @pytest.fixture(scope="class")
    def partial_test_data(self) -> dict[str, list[Any]] | pd.DataFrame:
        return {
            "marginals": [1, 1],
            "levels": [1, 3],
        }

    @pytest.fixture(scope="class")
    def remaining_test_data(self) -> dict[str, list[Any]] | pd.DataFrame:
        return {
            "marginals": [-2],
            "levels": [2],
        }

```

The tests are indeed not very documented and that needs work. I tried to write them all very clean and easy to read, but there are just... 2000 of them or something--so its a little challenging.

khaeru:
> This is a common pattern in the tests to write a base class that defines tests and parameterize them by implementing subclasses that implement the test data fixtures.

I've used a similar pattern myself elsewhere: https://github.com/khaeru/sdmx/blob/86e7bafe8944fe36c6ff2ac2889b287da6313bab/sdmx/tests/test_sources.py#L21-L101

> now I am looking at the file again it might make more sense to add another subclass of VariableDataTest which overrides the test_data_indexsets fixture in addition to implementing the other ones.

In the last commit on the branch, I renamed the existing `VariableDataTest.test_variable_add_data()` to `…add_data_2d()`, and added a similar method `…add_data_0d()`. I ran this and it passed, although I am not sure if the failure of other tests on some CI jobs is related/consequential to that.

Per your explanation, in data/test_optimization_variable_service.py, I see that we could instead leave VariableDataTest alone; maybe rename TestVariableData to TestVariableData2D, and add a TestVariableData0D similar to the snippet in your comment.

In core/test_optimization_variable.py, it looks like the fixture test_data_indexsets is defined on the intermediate class (VariableDataTest) rather than on the subclass that pytest will collect (TestVariableData). So I guess that would require a different approach, or perhaps the fixture could be moved to align with the other test module.

> Was able to quickly reproduce the original ValueError on a different branch with this test

This actually raises something maybe missed: if there are 0 index sets, there should be at most 1 value stored, right? Otherwise these:
```
"marginals": [-2, 1, 1],
"levels": [2, 1, 3],
```
…cannot be disambiguated in any way. IndexedRepository.validate_data() should raise an exception if there are 0 index sets and >1 data points.

If this is not correct then maybe I am missing something conceptually about the data model.

FWIW I will be on leave/travel until 1 October inclusive. I can come back to this then, but I would also be fine if you make whatever changes you feel are consistent with your own understanding of the data model and test suite design. Then I could study and understand those as expression of what I should do if I need to make future PRs.

PR #270, opened 2026-09-21
Enable `--workers` param for the server CLI

meksor:
Refactors the server object creation into a separate factory function since the workers argument does not work when directly supplying the asgi app as an object. I originally avoided this because it means all configuration from the cli command has to be written to the environment, but it ended up only being relevant for the `--debug` flag. 
The factory function is passed to uvicorn as an import string.


pmussak (approved):
lgtm

PR #271, opened 2026-09-21
Re-implement Upload Chunking

meksor:
In an attempt to simplify the code base I removed the upload chunking code previously. 
Users have been having issues with large uploads via HTTP, especially when there is contention on the service. 
Re-implementing upload chunking for large uploads (chunksize default: 10_000) should make that a little bit more stable.

This reactivates the env var `IXMP4_CLIENT__DEFAULT_UPLOAD_CHUNK_SIZE` and setting `Settings.client.default_upload_chunk_size`/`ClientSettings.default_upload_chunk_size`.
Also sets client concurrency to 1 per default and increases timeout to 60 seconds for further stability. 

Main logic and code for review are about 100 LOC in each of `ixmp4/data/services/procedure/chunking.py`
and `ixmp4/data/services/procedure/client.py`.

Chunking can be enabled in the procedure decorator call, for example: 
```py
class DataPointService(Service):
...
    @procedure(Http(methods=("POST",)), chunked=True) # or chunked="df"
    def bulk_upsert(self, df: SerializableDataFrame) -> None:
...
```
No safety/atomicity mechanisms exist here. If subsequent chunks fail, incomplete data may remain.


ixmp4/data/services/procedure/client.py
```diff
@@ -180,6 +185,127 @@ def dispatch_pagination_requests(
...
+    def chunk_payload_value(self, value: Any, chunk_size: int) -> list[Any]:
+        """Splits *value* into a list of serialized chunks.
+
+        Supports :class:`pandas.DataFrame` (and its serialized dict form)
+        as well as plain lists. Other payload types cannot be chunked.
+        """
+        if isinstance(value, pd.DataFrame):
+            df = value
+        elif isinstance(value, list):
+            return [value[i : i + chunk_size] for i in range(0, len(value), chunk_size)]
```
pmussak: I'd put this branch of the if-else block first, since it is the only one that includes an early return. I find early returns being placed as close to the start of a function/ method definition as possible easier to read.

ixmp4/data/services/procedure/client.py
```diff
@@ -180,6 +185,127 @@ def dispatch_pagination_requests(
...
+            for i in range(0, len(df), chunk_size)
+        ]
+
+    @staticmethod
+    def is_serialized_dataframe(value: Any) -> bool:
+        return (
+            isinstance(value, dict)
+            and "data" in value
+            and ("columns" in value or "dtypes" in value or "index" in value)
+        )
```
pmussak: Better to put this in `dataframe.py`, since it depends on the same contract of how dataframes are serialized that `serialize_df` implements.

ixmp4/conf/settings.py
```diff
@@ -53,12 +54,13 @@ class ClientSettings(BaseSettings):
     default_upload_chunk_size: int = Field(
         10_000,
```
pmussak, line 55: To prevent erroneous settings I'd restrict this to at least 1. Otherwise `[value[i : i + chunk_size]` in `chunk_payload_value` will produce unexpected results.
```suggestion
        10_000,
        ge=1
```

pmussak (changes requested):
Except for one recommendation for more defensive programming (restrict `default_upload_chunk_size` to >= 1) this looks good to me. Left a couple of nitpicks.

pmussak (approved):
lgtm

